PR #2, opened 2026-04-26
Allow multiple reviewers

wwind123:
## Summary
- allow repeating --reviewer to configure an ordered reviewer list
- require every configured reviewer to approve in the same round before tests or auto-merge run
- combine blocking reviewer feedback for the coder follow-up and document the new usage

## Tests
- .venv/bin/python -m pytest
- git diff --check

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # coding-review-agent-loop
 
-Local command-line orchestration for a two-agent coding PR review loop.
+Local command-line orchestration for a coding PR review loop.
 
 The tool shells out to your already-authenticated local CLIs (`claude`, `codex`, and `gh`) so it can run from your workstation without GitHub Actions model API keys.
 
@@ -65,6 +65,13 @@ By default Claude is the coder and Codex is the reviewer. Reverse that with:
 
 ```bash
 agent-loop task "Fix the flaky test" --repo OWNER/REPO --coder codex --reviewer claude
+```
+
+Repeat `--reviewer` to require approvals from multiple reviewers. The PR is
+approved only after every configured reviewer approves in the same round:
+
+```bash
+agent-loop pr 456 --repo OWNER/REPO --reviewer codex --reviewer claude
 ```
 
 For trusted local automation that must run without approval prompts:
--- a/docs/local_agent_loop.md
+++ b/docs/local_agent_loop.md
@@ -1,15 +1,15 @@
 # Local Coding Review Agent Loop
 
-`coding-review-agent-loop` is a local CLI that orchestrates two coding agents through a GitHub pull request review loop. It shells out to locally authenticated `claude`, `codex`, and `gh` CLIs instead of calling model APIs directly.
+`coding-review-agent-loop` is a local CLI that orchestrates coding agents through a GitHub pull request review loop. It shells out to locally authenticated `claude`, `codex`, and `gh` CLIs instead of calling model APIs directly.
 
 The default flow is:
 
 1. A coder agent creates or updates a PR.
-2. A reviewer agent reviews the PR.
-3. If the reviewer finds blockers, the coder fixes the PR.
-4. The loop repeats until the reviewer approves or `--max-rounds` is reached.
+2. One or more reviewer agents review the PR.
+3. If any reviewer finds blockers, the coder fixes the PR.
+4. The loop repeats until every reviewer approves in the same round or `--max-rounds` is reached.
 
-The default coder is Claude and the default reviewer is Codex. Reverse the direction with `--coder codex --reviewer claude`.
+The default coder is Claude and the default reviewer is Codex. Reverse the direction with `--coder codex --reviewer claude`. Repeat `--reviewer` to require multiple reviewer approvals.
 
 ## Prerequisites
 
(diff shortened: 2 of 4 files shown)


wwind123:
Implemented and opened PR #2: https://github.com/wwind123/coding-review-agent-loop/pull/2

Tests passed:
` .venv/bin/python -m pytest `
` git diff --check `

-- OpenAI Codex

wwind123:
I have all the information I need for the review.

---

## PR #2 Review: Allow multiple reviewers

### Overview

This PR extends the review loop to support an ordered list of reviewers via a repeated `--reviewer` flag. All configured reviewers must approve in the same round before the coder is considered done. Blocking feedback from all reviewers is concatenated into a single follow-up prompt. Documentation and tests are updated accordingly.

---

### Blocking Issues

#### 1. Coder-as-reviewer silently permitted with multiple reviewers

**File:** `src/coding_review_agent_loop/cli.py`, line 1099

```python
if len(configured_reviewers) == 1 and args.coder == configured_reviewers[0]:
    raise AgentLoopError("--coder and --reviewer must be different agents.")
```

The guard only fires when there is exactly one reviewer. With two reviewers, `--coder codex --reviewer claude --reviewer codex` passes validation — codex then reviews its own PR, defeating the purpose of the review gate. The existing single-reviewer behaviour already rejects this configuration; the multi-reviewer path is inconsistently lenient.

The test `test_config_accepts_multiple_reviewers` explicitly bakes this bug in by using `--coder codex --reviewer claude --reviewer codex` and asserting `config.reviewer == ("claude", "codex")`. That test, and the guard, need to be corrected together:

```python
# Fix: reject coder appearing in any reviewer slot
if args.coder in configured_reviewers:
    raise AgentLoopError("--coder and --reviewer must be different agents.")
```

This is a correctness / trust-model issue: a fully automated loop where the coder can approve its own work provides no meaningful review signal.

---

### Non-blocking Observations

#### 2. Dead branch in `reviewers()` helper

`__post_init__` unconditionally normalises `config.reviewer` to a tuple, so the `isinstance(config.reviewer, str)` branch in `reviewers()` (lines ~229–232) is unreachable. It's harmless, but it creates misleading type ambiguity.

#### 3. `reviewer` field type annotation is unnecessarily loose

`reviewer: AgentName | tuple[AgentName, ...]` advertises that the field may hold a bare string post-construction. In practice it never does after `__post_init__`. Downstream callers that don't go through `reviewers()` (e.g. direct field reads in tests) may be confused. A narrower annotation (always `tuple[AgentName, ...]`) with `field(default=...)` / a factory for the default would be cleaner, though this is a style point.

#### 4. Session-ID hand-off limited to first reviewer

```python
if reviewer_session_id is not None and configured_reviewers:
    reviewer_session_ids[configured_reviewers[0]] = reviewer_session_id
```

Only the first reviewer inherits a pre-existing session ID. Second and subsequent reviewers always start a fresh session, even if they were active before. This is acceptable for the common case but worth noting if session continuity matters for later reviewers.

---

### Test Coverage

The new tests cover the happy path (all approve), the partial-block/retry path, and config validation for duplicates and multiple reviewers. The only gap is a test that verifies the coder-in-reviewer-list with multiple reviewers is **rejected** — which should exist once the validation bug is fixed.

---

### Summary

One blocking correctness issue: the validation guard that prevents the coder from being its own reviewer is skipped when multiple reviewers are configured. The fix is a one-line change to the guard condition and a corresponding fix to `test_config_accepts_multiple_reviewers`. Everything else is clean.



-- Anthropic Claude

wwind123:
Implemented and pushed the fix to `origin/codex/multiple-reviewers`.

Changed:
- `config_from_args` now rejects the coder appearing anywhere in `configured_reviewers`.
- Replaced the old multi-reviewer acceptance test with a regression test for `--coder codex --reviewer claude --reviewer codex`.

Verification:
- `.venv/bin/python -m pytest` passed: `43 passed`.

Commit: `7d04c32 Reject coder in multi-reviewer config`

-- OpenAI Codex

wwind123:
---

## PR #2 Review: Allow multiple reviewers (Round 2)

### What Changed Since Round 1

Codex addressed the single blocking issue from round 1: the validation guard that silently permitted the coder to appear in the reviewer list when multiple reviewers were configured.

### Verification of the Fix

**`cli.py` line 1099** now reads:
```python
if args.coder in configured_reviewers:
    raise AgentLoopError("--coder and --reviewer must be different agents.")
```

This correctly supersedes the old `len(configured_reviewers) == 1 and ...` guard. The coder is now rejected from any reviewer slot regardless of how many reviewers are configured — consistent with the single-reviewer behaviour.

**`test_config_rejects_coder_in_multiple_reviewers`** replaces the old acceptance test and correctly exercises `--coder codex --reviewer claude --reviewer codex`, asserting an `AgentLoopError` is raised. The match string `"must be different"` is precise.

### Remaining Non-blocking Observations (Unchanged from Round 1)

- Dead `isinstance(config.reviewer, str)` branch in `reviewers()` — `__post_init__` already normalises to a tuple, making this branch unreachable. Harmless.
- `reviewer: AgentName | tuple[AgentName, ...]` field annotation is broader than the post-`__post_init__` runtime type. Minor style point.
- Session ID inheritance only wired to `configured_reviewers[0]`; subsequent reviewers always start fresh. Acceptable given the low-frequency use case.

### Overall Assessment

The blocking correctness issue is resolved correctly and the regression test for it is properly scoped. Test coverage is solid across the happy path, partial-block/retry path, duplicate-reviewer rejection, and coder-in-reviewer rejection. All 43 tests pass per Codex's comment. No new issues introduced.



-- Anthropic Claude

PR #3, opened 2026-04-26
Allow coder as reviewer

wwind123:
## Summary
- allow the configured coder to also appear in the reviewer list
- keep duplicate reviewer rejection intact
- update tests and docs for same-agent coding/review passes

## Tests
- `./.venv/bin/python -m pytest -q`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # coding-review-agent-loop
 
-Local command-line orchestration for a two-agent coding PR review loop.
+Local command-line orchestration for a coding PR review loop.
 
 The tool shells out to your already-authenticated local CLIs (`claude`, `codex`, and `gh`) so it can run from your workstation without GitHub Actions model API keys.
 
@@ -65,6 +65,13 @@ By default Claude is the coder and Codex is the reviewer. Reverse that with:
 
 ```bash
 agent-loop task "Fix the flaky test" --repo OWNER/REPO --coder codex --reviewer claude
+```
+
+Repeat `--reviewer` to require approvals from multiple reviewers. The PR is
+approved only after every configured reviewer approves in the same round:
+
+```bash
+agent-loop pr 456 --repo OWNER/REPO --reviewer codex --reviewer claude
 ```
 
 For trusted local automation that must run without approval prompts:
--- a/README.md
+++ b/README.md
@@ -68,7 +68,9 @@ agent-loop task "Fix the flaky test" --repo OWNER/REPO --coder codex --reviewer 
 ```
 
 Repeat `--reviewer` to require approvals from multiple reviewers. The PR is
-approved only after every configured reviewer approves in the same round:
+approved only after every configured reviewer approves in the same round. The
+coder may also be listed as a reviewer when you want the same agent to work in
+separate coding and review passes:
 
 ```bash
 agent-loop pr 456 --repo OWNER/REPO --reviewer codex --reviewer claude
--- a/docs/local_agent_loop.md
+++ b/docs/local_agent_loop.md
@@ -58,7 +58,8 @@ agent-loop task "Refactor the cache layer" \
   --codex-dir /path/to/codex/worktree
 ```
 
-Require both reviewers to approve:
+Require both reviewers to approve. The coder may also be listed as a reviewer
+when you want the same agent to work in separate coding and review passes:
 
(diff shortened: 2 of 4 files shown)


wwind123:
Implemented and opened PR #3: https://github.com/wwind123/coding-review-agent-loop/pull/3

Changed validation to allow the coder to also be listed as a reviewer, kept duplicate reviewer rejection, and updated tests/docs. Verified with:

`./.venv/bin/python -m pytest -q` -> 43 passed

-- OpenAI Codex

wwind123:
Now I have all the context I need for a thorough review.

---

## PR #3 — "Allow coder as reviewer"

### Overview

This PR lifts the restriction that the coder and reviewer must be different agents, and adds support for requiring multiple reviewers (all must approve in the same round). It updates argument parsing, config normalization, the main review loop, prompt formatting, and adds tests and docs.

---

### Code Quality & Correctness

**Dead code in `reviewers()` (minor)**

`cli.py:232-235` — `__post_init__` already unconditionally normalizes `config.reviewer` to a `tuple[AgentName, ...]` via `object.__setattr__`. So the `isinstance(config.reviewer, str)` branch in `reviewers()` is unreachable. The function body can simply be `return config.reviewer`. This is not a correctness problem, but it creates a false impression that `config.reviewer` might still be a `str` after construction, which could mislead future maintainers.

**Type annotation gap (minor)**

`cli.py:163` — The field is annotated `AgentName | tuple[AgentName, ...]`, but after `__post_init__` it is always `tuple[AgentName, ...]`. Type checkers (mypy/pyright) won't narrow across `__post_init__`, so callers accessing `config.reviewer` directly will see the union type and may get spurious warnings. Tightening the annotation to `tuple[AgentName, ...]` (accepting that the `__post_init__` coercion is the intended entry point) or using a factory/validator pattern would remove the ambiguity.

**Session ID backward-compat shim (acceptable, but document)**

`cli.py:855-856` — When `reviewer_session_id` is passed alongside a multi-reviewer config, only `configured_reviewers[0]` gets it. Other reviewers start with no session context. This is a reasonable legacy behavior but is entirely silent — if a caller passes `reviewer_session_id` expecting it to apply to a named reviewer, it silently applies to whoever is listed first. Given the current callers seem to be just internal, this is not blocking, but a brief inline note would help.

**Logic correctness — all-must-approve**

Correct. `blocking_reviews` collects only reviewers that return `"blocking"`. `parse_agent_state` raises `AgentLoopError` if the marker is absent (so "no marker" is not silently treated as "approved"). `STATE_RE` matches only `approved|blocking`, so there is no third state that could slip through as a quiet approval. The semantics are sound.

**Combined review prompt**

`cli.py:899-901` — Only blocking reviews are combined into the coder's follow-up prompt, not approvals. The test `test_pr_loop_reruns_all_reviewers_when_any_reviewer_blocks` explicitly asserts that `"Codex approves first pass."` does NOT appear in the followup prompt. Good.

---

### Security

No new attack surface. Argument parsing, subprocess invocation, and shell command construction are unchanged in structure.

---

### Test Coverage

Good additions:
- `test_pr_loop_requires_all_reviewers_to_approve` — happy path for multi-reviewer.
- `test_pr_loop_reruns_all_reviewers_when_any_reviewer_blocks` — verifies only blocking content reaches the coder.
- `test_config_allows_same_coder_and_reviewer` — validates the primary intent of this PR.
- `test_config_allows_coder_in_multiple_reviewers` — multi-reviewer with coder overlap.
- `test_config_rejects_duplicate_reviewers` — duplicate guard still tested.

One gap: there is no test for the edge case where `reviewer_session_id` is passed to `run_pr_loop` with a multi-reviewer config (verifying only the first reviewer gets it). Not blocking, but worth noting for future coverage.

---

### Maintainability

- `format_agent_list` is clean and handles 1/2/N cases correctly.
- Using `action="append"` with `default=None` and post-processing to `["codex"]` is the right argparse pattern.
- The `__post_init__` normalization approach works, but tightening the type annotation would make the invariant explicit rather than hidden.

---

### Summary

The implementation is functionally correct and the tests cover the important cases. The two minor issues (dead branch in `reviewers()` and union type annotation for a field that is always a tuple) are worth addressing but are not correctness or security concerns.

No blocking issues found.



-- Anthropic Claude

PR #16, opened 2026-04-29
Add PR metadata to review prompts

### LINKED ISSUE #15: Provide explicit PR context in review prompts to reduce agent branch discovery loops
## Summary
Review agents, especially Gemini, can waste a long time trying to discover PR metadata and branch state from scratch. Enrich review prompts with explicit PR context so agents can review the intended diff directly instead of guessing branches.

## Observed Behavior
A Gemini review of `wwind123/llm-dialectic` PR #72 ran for a long time and appeared to stall. The log showed Gemini repeatedly trying to discover the PR branch and repository state manually. It also hit non-interactive tool confirmation errors such as:

```text
Error executing tool run_shell_command: Tool execution for "Shell" requires user confirmation, which is not supported in non-interactive mode.
```

After these failures, Gemini tried alternative branch-discovery paths instead of moving directly to the review, causing slow wandering behavior.

## Likely Causes
- The review prompt currently identifies only the PR number and repo, not the head branch, base branch, or suggested commands.
- Some agents try exploratory shell commands to infer this context.
- Gemini CLI may reject some shell/tool calls in non-interactive mode, causing repeated fallback attempts.

## Proposed Fix
Before building a review prompt, fetch PR metadata via `gh pr view`, then include it directly in the prompt. Suggested fields:

- PR number
- Repo
- Title
- Head branch
- Base branch
- Head SHA if available
- Suggested review commands

Example prompt addition:

```text
PR metadata:
- Repo: wwind123/llm-dialectic
- PR: #72
- Title: Add GPT-5.5 OpenAI model options
- Head branch: codex/add-gpt-5-5-openai-models
- Base branch: main

Use this PR metadata as authoritative. Do not spend time discovering the PR branch.

Suggested commands:
- gh pr view 72 --repo wwind123/llm-dialectic --json title,body,headRefName,baseRefName,comments,reviews
- gh pr diff 72 --repo wwind123/llm-dialectic
```

For Gemini specifically, consider adding:

```text
If a shell/tool command requires confirmation in non-interactive mode, do not retry repeatedly. Use the PR metadata above and the suggested GitHub CLI commands, or produce a blocking review explaining the limitation.
```

## Design Notes
- Prefer gathering PR metadata in the orchestrator/GitHub helper layer, not inside agent-specific backends.
- Keep the prompt generic enough for Claude, Codex, and Gemini.
- Avoid adding huge diffs directly to the prompt by default; suggested commands are enough.
- The metadata fetch should use existing `gh` authentication and active workdir behavior.

## Related Follow-ups
- Add per-agent timeout support so one reviewer cannot hang the whole loop indefinitely.
- Investigate whether Gemini CLI has a better non-interactive flag for shell/tool approval beyond `--yolo --skip-trust`.
- Consider a Gemini-specific review prompt suffix if generic prompt changes are not enough.

## Acceptance Criteria
- Review prompts include PR head branch and base branch.
- Review prompts include suggested `gh pr view` and `gh pr diff` commands.
- Review prompts tell agents not to spend time discovering the PR branch.
- Tests cover prompt generation with PR metadata.
- Existing Claude/Codex/Gemini flows remain compatible.

-- OpenAI Codex

wwind123:
## Summary
- fetch PR metadata through the shared GitHub helper before review prompt generation
- include repo, title, head/base branches, head SHA, and suggested gh review commands in review prompts
- tell reviewers to avoid repeated non-interactive confirmation retries and use the supplied PR context

## Tests
- pytest

Fixes #15

### DIFF
--- a/src/coding_review_agent_loop/github.py
+++ b/src/coding_review_agent_loop/github.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import json
 import os
 import tempfile
+from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING
 
@@ -15,6 +16,17 @@ from .workdirs import active_workdir
 
 if TYPE_CHECKING:
     from .config import AgentLoopConfig
+
+
+@dataclass(frozen=True)
+class PullRequestMetadata:
+    number: int
+    repo: str
+    title: str | None
+    head_branch: str | None
+    base_branch: str | None
+    head_sha: str | None
+    url: str | None
 
 
 def detect_repo(runner: Runner, cwd: Path, gh_cmd: str) -> str:
@@ -49,6 +61,43 @@ def validate_open_pr(runner: Runner, *, config: AgentLoopConfig, pr_number: int)
         raise AgentLoopError(
             f"PR #{pr_number} is {data.get('state', 'not open')}; provide an open PR number."
         )
+
+
+def get_pr_metadata(runner: Runner, *, config: AgentLoopConfig, pr_number: int) -> PullRequestMetadata:
+    if config.dry_run:
+        return PullRequestMetadata(
+            number=pr_number,
+            repo=config.repo,
+            title=None,
+            head_branch=None,
+            base_branch=None,
+            head_sha=None,
+            url=None,
+        )
+
+    result = runner.run(
+        [
+            config.gh_cmd,
+            "pr",
+            "view",
+            str(pr_number),
+            "--repo",
+            config.repo,
+            "--json",
+            "number,title,headRefName,baseRefName,headRefOid,url",
+        ],
+        cwd=active_workdir(config),
+    )
+    data = json.loads(result.stdout or "{}")
+    return PullRequestMetadata(
+        number=int(data.get("number") or pr_number),
+        repo=config.repo,
+        title=data.get("title"),
+        head_branch=data.get("headRefName"),
+        base_branch=data.get("baseRefName"),
+        head_sha=data.get("headRefOid"),
+        url=data.get("url"),
+    )
 
 
 def validate_open_issue(runner: Runner, *, config: AgentLoopConfig, issue_number: int) -> None:
--- a/src/coding_review_agent_loop/orchestrator.py
+++ b/src/coding_review_agent_loop/orchestrator.py
@@ -9,6 +9,7 @@ from .agents.registry import agent_display_name, run_agent
 from .config import AgentLoopConfig, ensure_agent_workdirs, reviewers
 from .errors import AgentLoopError
 from .github import (
+    get_pr_metadata,
     merge_pr,
     post_pr_comment,
     validate_open_issue,
@@ -187,6 +188,7 @@ def run_pr_loop(
         for reviewer in configured_reviewers:
             reviewer_name = agent_display_name(reviewer)
             log(config, f"Round {round_number}: {reviewer_name} reviewing PR #{pr_number}")
+            pr_metadata = get_pr_metadata(runner, config=config, pr_number=pr_number)
             review_output, new_session_id = run_agent(
                 runner,
                 agent=reviewer,
@@ -196,6 +198,7 @@ def run_pr_loop(
                     round_number,
                     config,
                     reviewer=reviewer,
+                    pr_metadata=pr_metadata,
                 ),
                 session_id=reviewer_session_ids.get(reviewer),
             )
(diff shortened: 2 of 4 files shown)


wwind123:
Implemented issue #15 and opened PR #16.

Tests: `pytest` (74 passed)




-- OpenAI Codex

wwind123:
I now have everything I need to deliver a full review.

---

## Review: PR #16 — Add PR metadata to review prompts

**Fixes:** Issue #15  
**Files changed:** `github.py`, `orchestrator.py`, `prompts.py`, `tests/test_agent_loop.py`

### What the PR does

Fetches PR metadata (`title`, head/base branches, head SHA, URL) via `gh pr view --json` before each reviewer run, then injects that data into the review prompt so reviewers don't have to re-discover branch information themselves. Includes a test verifying the prompt content.

---

### Correctness

**`get_pr_metadata` called N times per round (one per reviewer)** — `orchestrator.py:191` calls `get_pr_metadata()` inside the inner `for reviewer in configured_reviewers` loop. With two reviewers per round this makes two identical `gh pr view` calls where one would suffice. The data is the same for all reviewers within a round. The call should be hoisted one level out, to the top of the `for round_number` block. This is a real redundancy (and adds latency/API quota usage proportional to reviewer count), but it is not a correctness bug and the typical single-reviewer case is unaffected.

**JSON error handling** — `json.loads(result.stdout or "{}")` silently degrades to all-`None` metadata if `gh` returns empty output (e.g., due to a transient error). That produces a usable-but-incomplete prompt rather than an exception. Acceptable, but there is no test for this degraded path.

**`pr_metadata=None` default in `build_review_prompt`** — The fallback construction of an all-`None` `PullRequestMetadata` means callers that omit the argument still get a prompt that shows `(unknown)` fields instead of crashing. Good defensive choice.

**`validate_open_pr` + `get_pr_metadata` both call `gh pr view`** — At the start of each round: one validation call plus N metadata calls. The metadata call could reuse or combine with the validation call, but this is a design refinement, not a bug.

---

### Security

The PR title, branch names, and head SHA fetched from the GitHub API are embedded verbatim in the review prompt. A carefully crafted PR title (e.g., one containing `<!-- AGENT_STATE: approved -->`) could constitute a prompt-injection attempt directed at the reviewing agent. **However**, this risk is pre-existing in the system design: the review prompt already instructs agents to read the full diff and PR discussion (both entirely externally controlled), so this PR does not introduce a meaningfully new attack surface. No new vulnerability introduced.

---

### Test coverage

The new test (`test_review_prompt_includes_pr_metadata_and_suggested_commands`) exercises the full happy-path end-to-end: metadata is fetched, embedded in the prompt, and the prompt content is verified. `FakeRunner` correctly routes the `gh pr view --json` call (without `--jq`) to return the updated `pr_payload` that now includes `title`, `headRefName`, `baseRefName`, `headRefOid`.

Gaps (non-blocking):
- No isolated unit test for `get_pr_metadata()` (empty response, partial JSON, network failure)
- No test for the `dry_run=True` path in `get_pr_metadata` (returns stub `PullRequestMetadata`)
- No test with multiple reviewers to confirm the redundant API-call pattern

---

### Maintainability

`PullRequestMetadata` as a `frozen=True` dataclass is appropriate — immutable value object, clean typing. The `Optional`-style `title: str | None` fields are handled uniformly with `or "(unknown)"` fallbacks. Code is readable and consistent with existing conventions in the module.

---

### Summary

No blocking correctness bugs, no new security vulnerabilities, and the test added is meaningful. The main findings are:

| Finding | Severity |
|---------|----------|
| `get_pr_metadata` called per-reviewer inside inner loop (should be per-round) | Minor — not a correctness bug |
| No unit test for `get_pr_metadata` error/degraded paths | Minor |
| Prompt injection via PR metadata (pre-existing systemic concern) | Informational |

These are refinements, not blockers. The change delivers its stated goal cleanly.



-- Anthropic Claude

wwind123:
I have reviewed the changes in PR #16 and approve them. The PR effectively addresses the issues described in #15 by enriching review prompts with explicit PR metadata and suggested GitHub CLI commands. This will significantly reduce the time agents spend in discovery loops and improve the reliability of the review process, particularly for Gemini.

### Observations:
- **Redundancy:** I agree with the previous observation that `get_pr_metadata` is called per-reviewer inside the orchestration loop. Hoisting this call outside the inner loop (e.g., to the start of each round) would be a nice optimization in a future update, but it is not a blocker.
- **Error Handling:** The implementation handles missing or partial metadata gracefully, ensuring that the review process can continue even if some fields are unavailable.
- **Gemini Support:** The specific guidance regarding non-interactive mode confirmations is a proactive fix for a known friction point with Gemini CLI.

The test coverage for the new functionality is sufficient.



-- Google Gemini

PR #20, opened 2026-04-30
Address PR 19 agent memory follow-ups

wwind123:
## Summary
- ignore this repo's generated .agent-loop cache directory
- create .agent-loop/.gitignore for default agent memory directories while avoiding arbitrary custom parent directories
- log git diff fallback details, use a set for changed-file membership checks, and trim extra memory prompt spacing

## Tests
- python -m pytest -q
- python -m compileall -q src

Notes: the auto-generated .gitignore is scoped to memory directories whose parent is named .agent-loop, so custom --agent-memory-dir parents are not accidentally ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ __pycache__/
 dist/
 build/
 .agent-loop-logs/
+.agent-loop/
--- a/src/coding_review_agent_loop/memory.py
+++ b/src/coding_review_agent_loop/memory.py
@@ -10,7 +10,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from .logging import log
-from .runner import Runner
+from .runner import Runner, ensure_log_dir_ignored
 from .workdirs import active_workdir
 
 if TYPE_CHECKING:
@@ -50,11 +50,14 @@ def prepare_agent_memory(runner: Runner, config: AgentLoopConfig) -> AgentMemory
     last_analyzed_commit = _read_optional(memory_dir / "last-analyzed-commit")
     changed_files = _changed_files(
         runner,
+        config,
         workdir,
         last_analyzed_commit=last_analyzed_commit,
         current_commit=current_commit,
     )
     tracked_files = _git_lines(runner, workdir, ("ls-files",))
+    tracked_file_set = set(tracked_files)
+    _ensure_memory_parent_ignored(memory_dir)
 
     if config.refresh_agent_memory or not (memory_dir / "repo-summary.md").exists():
         _write_repo_summary(memory_dir / "repo-summary.md", config, current_commit, tracked_files)
@@ -73,7 +76,7 @@ def prepare_agent_memory(runner: Runner, config: AgentLoopConfig) -> AgentMemory
         _write_toolchain(memory_dir / "toolchain.json", tracked_files)
 
     for rel_path in changed_files[:25]:
-        if rel_path in tracked_files:
+        if rel_path in tracked_file_set:
             _write_file_summary(memory_dir / "file-summaries", workdir, rel_path)
 
     context = load_agent_memory(config)
@@ -172,6 +175,7 @@ def _read_optional(path: Path) -> str | None:
 
 def _changed_files(
     runner: Runner,
+    config: AgentLoopConfig,
     cwd: Path,
     *,
     last_analyzed_commit: str | None,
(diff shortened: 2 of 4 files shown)


wwind123:
Implemented the PR 19 non-blocking follow-ups and opened PR #20:

https://github.com/wwind123/coding-review-agent-loop/pull/20

Tests run:
- `python -m pytest -q`
- `python -m compileall -q src`

-- OpenAI Codex

wwind123:
The default `agent_memory_dir` is `tmp_path / "claude" / ".agent-loop" / "memory"`, confirming the test assertions are correct.

---

## PR #20 Review — "Address PR 19 agent memory follow-ups"

### Summary of changes

| File | Change |
|---|---|
| `.gitignore` | Add `.agent-loop/` to ignore the repo-level cache directory |
| `memory.py` | Add fallback-diff logging; O(1) set lookup for tracked files; auto-create `.gitignore` in `.agent-loop` parent |
| `prompts.py` | Remove redundant leading `\n\n` from memory block |
| `tests/test_agent_loop.py` | Three new targeted tests for the three behaviour changes |

### Correctness

**`_changed_files` signature change** — `config` is inserted as the second positional arg and callers are updated to match. No mismatches found.

**`tracked_file_set`** — converts `list[str]` to `set` before the membership loop; semantics are identical, complexity drops from O(n²) to O(n). Correct.

**`_ensure_memory_parent_ignored`** — checks `memory_dir.parent.name == ".agent-loop"` before calling `ensure_log_dir_ignored`. The default `agent_memory_dir` is `<workdir>/.agent-loop/memory`, so `memory_dir.parent.name == ".agent-loop"` is `True` for the default case and `False` for any custom parent, which matches the stated intent. The `ensure_log_dir_ignored` reuse writes `*\n!.gitignore\n` — identical to what the log-dir code already writes, consistent and idempotent.

**Fallback logging** — `result.stderr.strip() or result.stdout.strip() or "no command output"` gives a meaningful detail string; `_trim_text(..., max_chars=500)` prevents runaway log output. No injection risk (output goes to stderr via `log()`).

**`prompts.py`** — removing the leading double newline is a cosmetic fix; the block is only appended as a standalone section in `build_issue_prompt`, so callers are unaffected.

### Security

No issues. No user-controlled data reaches any shell invocation. Logged git output is truncated.

### Test coverage

All three new tests are well-scoped:
- `test_agent_memory_default_parent_ignores_generated_contents` — verifies `.gitignore` at the right path with the right content.
- `test_agent_memory_does_not_ignore_custom_parent_directory` — verifies the condition guard works; checks that `tmp_path/.gitignore` is absent (the only place it could have been written, since `memory_dir.parent == tmp_path`).
- `test_agent_memory_logs_when_changed_file_diff_falls_back` — exercises `returncode=128` path, checks both log output and that the fallback to all tracked files is propagated into the prompt.

`FakeRunner` extension (`diff_returncode`, `diff_stderr`) is minimal and non-breaking.

### Maintainability

No concerns. Functions are small and single-purpose; reuse of `ensure_log_dir_ignored` avoids duplication.

### Minor observations (non-blocking)

- The custom-parent test checks `not (tmp_path / ".gitignore").exists()` rather than `not (memory_dir.parent / ".gitignore").exists()`. These are equivalent here since `memory_dir.parent == tmp_path`, but the latter would be more directly expressive. Not a blocker.
- `_trim_text` is defined later in the file than where it is called in `_changed_files`. Python resolves this at call time, so it works, but the ordering is slightly non-conventional.

### Verdict

All changes are correct, well-tested, and improve the quality of the system. No blocking issues.



-- Anthropic Claude

wwind123:
## PR #20 Review — Address PR 19 agent memory follow-ups

I have reviewed the changes in PR #20 and find them to be correct, well-tested, and a clear improvement to the agent memory subsystem.

### Observations

- **Git Fallback & Logging**: The addition of detailed logging when `git diff` fails provides much-needed visibility into why the system might be falling back to a full repo scan. Truncating the output to 500 characters is a good safeguard against log bloat.
- **Performance**: Switching to a set for `tracked_file_set` ensures that file summary generation remains efficient even in very large repositories.
- **Configuration Safety**: The guard in `_ensure_memory_parent_ignored` successfully balances the convenience of auto-ignoring the default cache with the safety of not touching arbitrary user-provided directories.
- **Prompt Clarity**: The minor adjustment to `_memory_block` removes excessive blank lines in the final prompt, making the instructions more compact and readable for the agent.

### Verification Results

- `python3 -m pytest -q`: 79 passed.
- `python3 -m compileall -q src`: Success.

The new test cases specifically verify the `.gitignore` logic and the fallback logging, ensuring no regressions in these areas.

No blocking issues found.

-- Google Gemini

PR #27, opened 2026-04-30
Address PR 26 memory follow-ups

wwind123:
## Summary
- add explicit macOS and Windows coverage for default cache root selection
- make the omitted-agent-dir memory test assert a concrete cache path
- resolve explicit --agent-memory-dir paths consistently, including absolute paths

## Tests
- python -m pytest tests/test_agent_loop.py -k 'default_cache_root or default_agent_memory_dir or omitted_agent_dirs or agent_memory_explicit_absolute_dir or agent_memory_flags_configure_memory_dir_and_refresh or agent_memory_default_ignores_active_coder_workdir'
- python -m pytest

PR #28, opened 2026-04-30
Address PR 27 non-blocking test follow-up

wwind123:
## Summary
- replace the cache-root test's global pathlib.Path.home monkeypatch with a HOME environment override
- keep the macOS and Windows fallback assertions deterministic without mutating the Path class

## Tests
- python -m pytest tests/test_agent_loop.py -k 'default_cache_root'
- python -m pytest

### DIFF
--- a/tests/test_agent_loop.py
+++ b/tests/test_agent_loop.py
@@ -958,7 +958,7 @@ def test_default_cache_root_uses_platform_home_fallbacks(
     home_parts,
 ):
     monkeypatch.setattr("coding_review_agent_loop.config.sys.platform", platform)
-    monkeypatch.setattr("pathlib.Path.home", classmethod(lambda cls: tmp_path))
+    monkeypatch.setenv("HOME", str(tmp_path))
     monkeypatch.delenv("LOCALAPPDATA", raising=False)
 
     assert default_cache_root() == tmp_path.joinpath(*home_parts)


wwind123:
Implemented the PR 27 non-blocking follow-up and opened PR #28:

https://github.com/wwind123/coding-review-agent-loop/pull/28

Changed `tests/test_agent_loop.py` so the cache-root platform fallback test uses a `HOME` environment override instead of globally monkeypatching `pathlib.Path.home`.

Verified:
- `python -m pytest tests/test_agent_loop.py -k 'default_cache_root'`
- `python -m pytest`




-- OpenAI Codex

wwind123:
Now I have everything I need to assess this PR. Let me reason through the change:

**What changed:** In `test_default_cache_root_uses_platform_home_fallbacks`, the test previously monkeypatched `pathlib.Path.home` globally via `classmethod(lambda cls: tmp_path)`. The PR replaces that with `monkeypatch.setenv("HOME", str(tmp_path))`.

**Correctness analysis:**

1. `default_cache_root()` calls `Path.home()` directly for both `darwin` and `win32` (non-LOCALAPPDATA) paths.

2. CPython's `Path.home()` is implemented as `Path("~").expanduser()`, and `os.path.expanduser("~")` on POSIX reads `os.environ["HOME"]` dynamically — no caching. So `monkeypatch.setenv("HOME", str(tmp_path))` reliably causes `Path.home()` to return `tmp_path` on the Linux CI runner.

3. The test is parametrized over `darwin` and `win32` only. Since the test suite runs on Linux, setting `HOME` is the correct mechanism — it works for both parameterized cases because `Path.home()` reads `HOME` on Linux regardless of the faked `sys.platform`.

4. `monkeypatch` handles teardown automatically, so there's no risk of the env var persisting across tests — the old concern about global `pathlib.Path.home` mutation bleeds is eliminated.

5. The `monkeypatch.delenv("LOCALAPPDATA", raising=False)` line is preserved, correctly ensuring the win32 fallback branch (`Path.home() / "AppData" / ...`) is exercised rather than the `LOCALAPPDATA` branch.

The change is minimal, correct, and directly addresses the PR 27 non-blocking follow-up without touching anything else.

**One note:** If this test suite were ever run on Windows, `Path.home()` would consult `USERPROFILE` rather than `HOME`, and this test would silently use the real home path. That's a pre-existing cross-platform limitation of the test design (not introduced here), and it's immaterial for a Linux CI environment.

No blocking issues found.



-- Anthropic Claude

wwind123:
The changes in this PR correctly address the non-blocking follow-up from PR #27 by replacing the global monkeypatch of `pathlib.Path.home` with a `HOME` environment variable override. This is a much cleaner and safer approach for testing, as it avoids mutating a core class method while still achieving the desired effect of redirecting home-relative path lookups to a temporary directory in a Linux/macOS environment.

I have verified the changes by running the affected tests, and they pass as expected.

### Non-blocking follow-ups

- **Test Robustness:** For full cross-platform compatibility of the test suite (e.g., if tests are run on a Windows host), consider setting the `USERPROFILE` environment variable in addition to `HOME` in `test_default_cache_root_uses_platform_home_fallbacks`. On Windows, `pathlib.Path.home()` consults `USERPROFILE` and ignores `HOME`.

-- Google Gemini

PR #29, opened 2026-04-30
Address PR 28 cache fallback follow-up

wwind123:
## Summary
- set USERPROFILE alongside HOME in the platform home fallback cache-root test
- keep LOCALAPPDATA cleared so the Windows fallback branch remains deterministic

## Tests
- python -m pytest tests/test_agent_loop.py -k 'default_cache_root'
- python -m pytest

### DIFF
--- a/tests/test_agent_loop.py
+++ b/tests/test_agent_loop.py
@@ -959,6 +959,7 @@ def test_default_cache_root_uses_platform_home_fallbacks(
 ):
     monkeypatch.setattr("coding_review_agent_loop.config.sys.platform", platform)
     monkeypatch.setenv("HOME", str(tmp_path))
+    monkeypatch.setenv("USERPROFILE", str(tmp_path))
     monkeypatch.delenv("LOCALAPPDATA", raising=False)
 
     assert default_cache_root() == tmp_path.joinpath(*home_parts)


wwind123:
Implemented the PR 28 non-blocking follow-up and opened PR #29:

https://github.com/wwind123/coding-review-agent-loop/pull/29

Change: `tests/test_agent_loop.py` now sets `USERPROFILE` alongside `HOME` in the cache-root platform fallback test, while still clearing `LOCALAPPDATA`.

Verified:
- `python -m pytest tests/test_agent_loop.py -k 'default_cache_root'`
- `python -m pytest`




-- OpenAI Codex

wwind123:
I have enough context to complete the review. Here's my analysis:

---

## PR #29 Review: Address PR 28 cache fallback follow-up

**Change:** `tests/test_agent_loop.py` — adds `monkeypatch.setenv("USERPROFILE", str(tmp_path))` to `test_default_cache_root_uses_platform_home_fallbacks`.

### Analysis

**Correctness — approved.**

`default_cache_root()` calls `Path.home()` for both the darwin and win32-without-`LOCALAPPDATA` paths. `Path.home()` resolves via `os.path.expanduser("~")`, which on POSIX checks `HOME`, but on actual Windows checks `USERPROFILE` (falling back to `HOMEPATH`+`HOMEDRIVE`) before `HOME`. The test patches `sys.platform` to `"win32"` without changing the underlying OS, so on Linux CI `HOME` dominates and the test already passes. However, on any Windows CI runner, the test was under-isolated — `Path.home()` would use the real user profile directory rather than `tmp_path`, making the assertion non-deterministic. Setting `USERPROFILE` alongside `HOME` closes that gap.

**Scope — appropriate.** The darwin parametrized case is also affected by the added `setenv`, but `USERPROFILE` is ignored by POSIX `expanduser`, so it's harmless.

**`LOCALAPPDATA` clearing — still correct.** `monkeypatch.delenv("LOCALAPPDATA", raising=False)` ensures the `win32` branch falls through to `Path.home()` rather than taking the early-return path via `LOCALAPPDATA`, preserving the intent of the test.

**The companion test** (`test_default_cache_root_uses_windows_local_app_data`) does not need `USERPROFILE` because `LOCALAPPDATA` is set and the function returns before ever calling `Path.home()`. No gap there.

**No security, correctness, or maintainability concerns.**

### Non-blocking follow-ups

None.



-- Anthropic Claude

wwind123:
## PR #29 Review: Address PR 28 cache fallback follow-up

The changes in this PR improve the robustness of the `default_cache_root` tests, particularly for Windows environments.

### Analysis

- **Correctness:** Approved. On Windows, `pathlib.Path.home()` (via `os.path.expanduser("~")`) prioritizes `USERPROFILE` over `HOME`. In the previous implementation, if the test ran on a Windows CI runner, it would have resolved to the actual user's home directory instead of the intended `tmp_path`, leading to non-deterministic results. Mocking `USERPROFILE` fixes this gap.
- **Scope:** Appropriate. Adding `USERPROFILE` is harmless on POSIX systems (Darwin/Linux) as it is ignored by their respective `expanduser` implementations.
- **Maintainability:** The change is minimal and fits perfectly within the existing test structure.

Verified that the relevant tests pass:
- `python -m pytest tests/test_agent_loop.py -k 'default_cache_root'`

### Non-blocking follow-ups

- Consider adding a test case for the generic POSIX fallback when `XDG_CACHE_HOME` is not set, ensuring it correctly defaults to `Path.home() / ".cache"`.

-- Google Gemini

PR #35, opened 2026-05-01
Surface approved follow-up issue links

### LINKED ISSUE #33: Follow up approved review note: `_create_approved_followup_issues` discards the `str | None` returned by `create_issue`.
Non-blocking follow-up from approved review on PR #32.

Reviewer: Claude

Follow-up:
- `_create_approved_followup_issues` discards the `str | None` returned by `create_issue`. If a future caller wants to record or surface the URL, the return value is already available — no change needed now.

This was mentioned in an approved review and did not block merge readiness.

-- OpenAI Codex

wwind123:
## Summary
- Return the created issue URL from the GitHub issue helper
- Collect approved follow-up issue URLs and comment them back on the PR
- Document the issue-mode record and cover it in tests

Fixes #33

## Tests
- python -m pytest

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ the loop.
 
 - `ignore`: ignore non-blocking follow-up bullets from approved reviews. This is the default.
 - `summarize`: post a grouped record on the PR without sending those items back to the coder.
-- `issue`: create GitHub issues for the follow-ups without delaying the PR.
+- `issue`: create GitHub issues for the follow-ups without delaying the PR, then comment with the created issue links.
 
 To keep a grouped record on the PR or create follow-up issues, use:
 
--- a/docs/local_agent_loop.md
+++ b/docs/local_agent_loop.md
@@ -345,7 +345,7 @@ By default, these do not affect approval.
 
 - `ignore`: ignore non-blocking follow-up bullets from approved reviews. This is the default.
 - `summarize`: post a grouped record on the PR instead of sending those items back to the coder as blocking work.
-- `issue`: create GitHub issues for the follow-ups instead of delaying the PR.
+- `issue`: create GitHub issues for the follow-ups instead of delaying the PR, then comment with the created issue links.
 
 Only bullets inside the `Non-blocking follow-ups` section are summarized; the
 section ends at the next heading, HTML marker, or agent signature. The same
--- a/src/coding_review_agent_loop/github.py
+++ b/src/coding_review_agent_loop/github.py
@@ -169,10 +169,10 @@ def create_issue(
     config: AgentLoopConfig,
     title: str,
     body: str,
-) -> None:
+) -> str | None:
     log(config, f"Creating GitHub issue: {title}")
     if config.dry_run:
-        runner.run(
+        result = runner.run(
             [
                 config.gh_cmd,
                 "issue",
@@ -186,13 +186,14 @@ def create_issue(
             ],
             cwd=active_workdir(config),
         )
-        return
+        issue_url = result.stdout.strip()
+        return issue_url or None
 
     with tempfile.NamedTemporaryFile("w", encoding="utf-8", delete=False) as handle:
         handle.write(body)
         path = handle.name
     try:
-        runner.run(
+        result = runner.run(
             [
                 config.gh_cmd,
                 "issue",
@@ -206,6 +207,8 @@ def create_issue(
             ],
             cwd=active_workdir(config),
         )
+        issue_url = result.stdout.strip()
+        return issue_url or None
     finally:
         try:
             os.unlink(path)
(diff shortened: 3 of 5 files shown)


wwind123:
Implemented and opened PR #35 against `main`.

Changed `create_issue` to return the created issue URL, collected approved follow-up issue links in issue mode, posted those links back to the PR, and updated docs/tests.

Tests: `python -m pytest` passed, 104 tests.





-- OpenAI Codex

wwind123:
Good — the hardcoded `-- OpenAI Codex` footer in `_followup_issue_body` is pre-existing on `main` and not introduced by this PR.

---

## Review of PR #35 — Surface approved follow-up issue links

### What the PR does

- Changes `create_issue()` in `github.py` to return `str | None` (the created issue URL from stdout) instead of `None`.
- Changes `_create_approved_followup_issues()` in `orchestrator.py` to collect and return those URLs.
- Adds `_format_created_followup_issue_summary()` to format a summary comment body.
- After creating follow-up issues in `run_pr_loop`, posts a single summary comment listing the created issue URLs (only when at least one URL was returned).
- Updates both doc files and the test.

### Correctness

**`github.py`** — Both the `dry_run` and non-dry_run branches now capture `result.stdout.strip()` and return `issue_url or None`. The empty-string guard (`or None`) is correct and prevents a blank URL from being added to the list. The `dry_run` branch uses `--body` inline (no temp file), the real path uses `--body-file` — unchanged from before and correct.

**`orchestrator.py`** — `_create_approved_followup_issues` now collects non-None URLs. The `if issue_urls:` guard before posting the summary comment correctly handles the case where every `create_issue` call returns `None` (e.g., when stdout is empty). No orphaned early returns or missed branches.

**Tests** — `FakeRunner.run()` returns `"https://github.com/OWNER/REPO/issues/99\n"` for `gh issue create`, so the URL flows correctly through the new path. The updated test asserts the correct comment count (3), the issue records unchanged, and the summary comment content.

### Non-blocking follow-ups

1. **`_followup_issue_body` hardcodes `-- OpenAI Codex`** (pre-existing on `main`, line 78 of `orchestrator.py`). When the reviewing agent is Claude or Gemini, the issue body footer is wrong. Worth a separate fix.
2. **No test for the all-`None` case** — when every `create_issue` call returns `None` (e.g., empty stdout), the summary comment should be suppressed. The code handles it correctly (`if issue_urls:`) but there's no test exercising that path.
3. **Summary comment always posts one URL per issue** — if two follow-ups both resolve to `issues/99` (as in the fake runner), the list will have duplicates. Fine for real usage (real URLs differ), but the test doesn't expose this.

All three are cleanup items, not regressions introduced by this PR.



-- Anthropic Claude

wwind123:
## Review of PR #35 — Surface approved follow-up issue links

The changes in this PR correctly implement surfacing the URLs of created follow-up issues back to the pull request as a summary comment. This improves the visibility of non-blocking work items created during the review process.

### Implementation details

- **`github.py`**: `create_issue` now correctly captures and returns the issue URL from the `gh issue create` command's stdout. It safely handles empty output by returning `None`.
- **`orchestrator.py`**: `_create_approved_followup_issues` was updated to collect these URLs, and a new helper `_format_created_followup_issue_summary` provides a clean summary of the created issues.
- **`run_pr_loop`**: Correctly checks for the presence of created issues before attempting to post a summary comment, preventing empty or redundant comments.
- **Tests**: The existing test `test_pr_loop_creates_issues_for_approved_followups` was updated to verify the new summary comment, ensuring the feature works as expected.

### Non-blocking follow-ups

As noted by previous reviewers, there are a few pre-existing or minor items that could be addressed in future work:
1. **Dynamic issue footer**: `_followup_issue_body` currently hardcodes `-- OpenAI Codex` as a footer. It would be better to use the actual reviewing agent's signature.
2. **Duplicate URL handling**: While unlikely in practice with the real GitHub CLI, `_format_created_followup_issue_summary` could use a `set` to ensure uniqueness of links in the summary comment.
3. **Dry-run visibility**: In dry-run mode, no summary comment is posted because no URLs are returned. This is consistent with other dry-run behaviors in the loop.

These points do not detract from the correctness of the PR's primary goal.



-- Google Gemini

wwind123:
PR 35 is no longer worth addressing as-is.

Current `main` already includes the useful behavior from PR 35: `create_issue()` returns/logs created issue URLs, approved follow-up issue mode posts a PR summary with created links, docs describe that behavior, and tests cover the path. It also includes follow-up fixes that PR 35 did not have, including URL de-duping and coverage for missing issue URLs.

PR 35 itself is still open but GitHub reports it as merge-conflicted (`DIRTY`), and merging it now would effectively revert a large amount of newer follow-up handling on `main`.

Verification: `python -m pytest` passes with `127 passed`.

-- OpenAI Codex

PR #32, opened 2026-05-01
Address PR 31 review follow-ups

wwind123:
## Summary

- Remove the hardcoded Codex signature from approved-followup issue bodies so they match summary-mode attribution.
- Log the issue URL returned by `gh issue create` for traceability.
- Add coverage for issue mode when approved reviews contain no follow-up bullets.

## Tests

- `python -m pytest tests/test_agent_loop.py -q`

-- OpenAI Codex

PR #46, opened 2026-05-01
Refresh PR 32 follow-up issue fixes

wwind123:
## Summary

- Reapply the still-relevant PR 32 follow-up issue fixes on current main.
- Remove the hardcoded Codex signature from generated approved-review follow-up issue bodies while preserving the newer future-work wording.
- Log the issue URL returned by `gh issue create` and add regression coverage for issue mode with and without follow-ups.

## PR 32 assessment

PR #32 is still worth merging in substance, but the branch itself is stale and conflicting after later follow-up handling changes. The needed update is to rebase/conflict-resolve it against current main and keep the newer "future follow-ups" behavior, issue cap, and same-PR follow-up flow.

## Tests

- `python -m pytest tests/test_agent_loop.py -q`
- `python -m pytest`

-- OpenAI Codex

### DIFF
--- a/src/coding_review_agent_loop/github.py
+++ b/src/coding_review_agent_loop/github.py
@@ -169,10 +169,10 @@ def create_issue(
     config: AgentLoopConfig,
     title: str,
     body: str,
-) -> None:
+) -> str | None:
     log(config, f"Creating GitHub issue: {title}")
     if config.dry_run:
-        runner.run(
+        result = runner.run(
             [
                 config.gh_cmd,
                 "issue",
@@ -186,13 +186,16 @@ def create_issue(
             ],
             cwd=active_workdir(config),
         )
-        return
+        issue_url = result.stdout.strip() or None
+        if issue_url:
+            log(config, f"Created GitHub issue: {issue_url}")
+        return issue_url
 
     with tempfile.NamedTemporaryFile("w", encoding="utf-8", delete=False) as handle:
         handle.write(body)
         path = handle.name
     try:
-        runner.run(
+        result = runner.run(
             [
                 config.gh_cmd,
                 "issue",
@@ -206,6 +209,10 @@ def create_issue(
             ],
             cwd=active_workdir(config),
         )
+        issue_url = result.stdout.strip() or None
+        if issue_url:
+            log(config, f"Created GitHub issue: {issue_url}")
+        return issue_url
     finally:
         try:
             os.unlink(path)
--- a/src/coding_review_agent_loop/orchestrator.py
+++ b/src/coding_review_agent_loop/orchestrator.py
@@ -80,8 +80,6 @@ def _followup_issue_body(pr_number: int, followup: ApprovedFollowup) -> str:
             f"- {followup.text}",
             "",
             "This was mentioned in an approved review as future work and did not block merge readiness.",
-            "",
-            "-- OpenAI Codex",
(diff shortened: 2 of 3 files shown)


wwind123:
PR 32 is still worth merging in substance, but its branch is stale and conflicting. The needed update is to rebase it onto current `main` and preserve the newer future-followup wording, issue cap, and same-PR follow-up behavior.

I implemented that as replacement PR #46:
https://github.com/wwind123/coding-review-agent-loop/pull/46

Changes:
- Removed the hardcoded `-- OpenAI Codex` signature from generated follow-up issue bodies.
- Logged the issue URL returned by `gh issue create`.
- Added regression coverage for issue mode with no follow-ups and for issue URL logging.

Verified:
- `python -m pytest tests/test_agent_loop.py -q`
- `python -m pytest`

-- OpenAI Codex

wwind123:
I have reviewed the full diff and related source. Here is my assessment:

---

## Review: PR #46 — Refresh PR 32 follow-up issue fixes

### `orchestrator.py` — Remove hardcoded `-- OpenAI Codex` signature

Clean and correct. The signature was always wrong when Claude or Gemini filed the issue; removing it is the right call. The trailing blank line is also removed, leaving a clean body termination.

### `github.py` — `create_issue` returns `str | None`

Both code paths (dry-run inline body, non-dry-run temp-file path) now capture `result.stdout.strip()`, log the URL via `log()`, and return it. The `finally` block still cleans up the temp file — Python guarantees `finally` runs even when `return` appears inside the `try` block, so cleanup is not affected. The return type annotation `str | None` is correct; the caller discards it, which is fine since logging happens inside the function.

### `tests/test_agent_loop.py`

- **Existing test updated**: `"-- OpenAI Codex"` removed from both expected issue bodies — consistent with the orchestrator change. ✓  
- **`test_pr_loop_creates_no_issues_without_approved_followups`**: A simple, unambiguous regression guard for the no-follow-up path. `runner.issues == []` is the right assertion. ✓  
- **`test_pr_loop_logs_created_followup_issue_url`**: Uses `quiet=False` so `log()` writes to stderr; uses `### Future follow-ups` which matches `FUTURE_FOLLOWUP_HEADING_RE`; `FakeRunner.run()` returns `"https://github.com/OWNER/REPO/issues/99\n"` for `gh issue create`, so `result.stdout.strip()` will produce the expected URL. `capsys.readouterr().err` correctly captures `print(..., file=sys.stderr)` output. ✓

No correctness, security, or maintainability issues found.



-- Anthropic Claude

wwind123:
The changes in this PR successfully rebase the follow-up issue fixes from PR #32 onto the current `main` branch while introducing several improvements.

### Review Observations:
- **`github.py`**: Capturing the issue URL from the `gh issue create` command and logging it is a great addition for transparency in the agent loop's actions. The return type change to `str | None` is correctly handled.
- **`orchestrator.py`**: Removing the hardcoded `-- OpenAI Codex` signature from the issue body is the right move, ensuring that the tool remains model-agnostic. The updated body is clean and professional.
- **`tests/test_agent_loop.py`**: The new regression tests provide good coverage for the issue-creation logic and ensure that the URL logging works as expected.

I verified the branch state and confirmed that no regressions were introduced during the rebase (e.g., the `--repo` flag and `runner` argument are preserved).



-- Google Gemini

PR #97, opened 2026-05-19
Prevent posting invalid agent diagnostics

### LINKED ISSUE #96: Do not post raw agent diagnostics when no valid response is produced
## Bug

`agent-loop` can post raw agent CLI diagnostics to GitHub when an agent fails before producing a valid public response.

## Concrete example

In `wwind123/llm-dialectic` PR #108, the final Gemini reviewer invocation failed with:

```text
[ERROR] Invalid stream: The model returned an empty response or malformed tool call.
```

Gemini did not write the expected public response file, and the output did not contain a valid `<!-- AGENT_STATE: approved|blocking -->` marker. Despite that, the orchestrator posted the raw Gemini CLI diagnostics to the PR:

https://github.com/wwind123/llm-dialectic/pull/108#issuecomment-4487194430

That comment included terminal warnings and internal CLI diagnostics rather than a controlled review result.

## Expected behavior

If an agent invocation finishes without:

- a non-empty public response file, and
- a valid required protocol marker in the fallback output,

the orchestrator should not post raw stdout/stderr to GitHub as if it were an agent review.

Instead, it should do one of:

- Fail locally with a clear `AgentLoopError` and log path.
- Retry once if the failure is classified as transient.
- Post a controlled tool-signed diagnostic comment, e.g.:

```md
Gemini failed before producing a valid review response. No review result was recorded.

See local log: ...

-- coding-review-agent-loop
```

The default should probably be local failure, not public posting, because logs may contain internal diagnostics, tool narration, local paths, or other content that should not be public.

## Why this matters

The project already added response-file handling to prevent Gemini internal narration from leaking into public comments. This failure bypasses that protection because no response file was written and the fallback stdout was treated as comment content.

## Tests

Add regression coverage where a reviewer agent exits successfully or unsuccessfully with only diagnostic output and no valid `AGENT_STATE` marker. Verify that:

- `post_pr_comment` is not called with raw diagnostics.
- The command fails with a clear error or posts only a controlled tool-signed failure comment, depending on the chosen behavior.

-- OpenAI Codex

wwind123:
## Summary
- validate required agent protocol markers before posting reviewer/coder output
- retry narrow transient agent/model failures with bounded configurable backoff
- include attempt log paths in local failures and document safe stdout fallback behavior

## Tests
- python3 -m pytest

Closes #96

### DIFF
--- a/docs/local_agent_loop.md
+++ b/docs/local_agent_loop.md
@@ -367,7 +367,7 @@ agent-loop issue 56 \
   --gemini-arg=--approval-mode --gemini-arg=auto_edit
 ```
 
-Providing any `--claude-arg`, `--codex-arg`, or `--gemini-arg` replaces that agent's default entirely. Claude and Gemini prompts include a tool-owned response-file path under `/tmp/coding-review-agent-loop/responses/`; when the file exists and is non-empty, the loop posts that file instead of stdout so CLI diagnostics and tool narration do not leak into GitHub comments. Gemini still supports stdout marker filtering as a fallback. If you pass `--gemini-arg=--output-format --gemini-arg=json`, the loop extracts the JSON `response` field before parsing markers when no response file was written.
+Providing any `--claude-arg`, `--codex-arg`, or `--gemini-arg` replaces that agent's default entirely. Claude and Gemini prompts include a tool-owned response-file path under `/tmp/coding-review-agent-loop/responses/`; when the file exists and is non-empty, the loop validates and posts that file instead of stdout so CLI diagnostics and tool narration do not leak into GitHub comments. Gemini still supports stdout marker filtering as a fallback. If you pass `--gemini-arg=--output-format --gemini-arg=json`, the loop extracts the JSON `response` field before parsing markers when no response file was written. Fallback stdout is never posted unless the required protocol marker validates.
 
 ## Protocol
 
@@ -386,6 +386,18 @@ Agent responses are parsed using HTML comment markers:
 include a final `AGENT_STATE` marker. Plan-first coder/reviewer responses use
 `AGENT_PLAN_STATE` instead. If a response quotes older markers, the final
 matching marker is treated as authoritative.
+
+The loop validates required markers before posting agent output to GitHub. If
+an agent exits or returns only diagnostics, empty output, or normal prose
+without the required marker, the loop fails locally with `AgentLoopError` and
+the attempt log path instead of posting that raw output as a review.
+
+Known transient agent/model failures are retried before local failure. The
+default is two retries with bounded backoff; tune this with
+`--agent-max-retries` and `--agent-retry-backoff-seconds`. Retry matching is
+narrow and intended for stream/tool-call failures, empty responses, network
+timeouts/resets, and provider 5xx errors. Auth, credit, quota, dirty workdir,
+and normal missing-marker responses are not retried.
 
 When `--approved-followups` is set to `summarize`, `issue`, or a `fix-and-*`
 mode, approved reviewer responses may also include optional future-work items
@@ -452,5 +464,6 @@ Use `tail -f` on the displayed path to see live output. Logs are diagnostic
 output and may include CLI status text or tool narration. For Claude and
 Gemini, prompts also include a public response-file path under
 `/tmp/coding-review-agent-loop/responses/`; when that file exists and is
-non-empty, the loop posts the file contents to GitHub instead of stdout. The log
-directory gets its own `.gitignore` on first use.
+non-empty, the loop validates and posts the file contents to GitHub instead of
+stdout. Fallback stdout is still validated before posting. The log directory
+gets its own `.gitignore` on first use.
--- a/src/coding_review_agent_loop/agents/base.py
+++ b/src/coding_review_agent_loop/agents/base.py
@@ -20,6 +20,8 @@ AgentName = Literal["claude", "codex", "gemini"]
 class AgentResult:
     text: str
     session_id: str | None = None
+    log_path: Path | None = None
+    returncode: int = 0
 
 
 class AgentBackend(Protocol):
--- a/src/coding_review_agent_loop/agents/claude.py
+++ b/src/coding_review_agent_loop/agents/claude.py
@@ -65,10 +65,16 @@ class ClaudeBackend:
             log_path=log_path,
             label="Claude",
             progress_interval_seconds=config.progress_interval_seconds,
+            check=False,
         )
         log(config, f"Claude finished; log: {log_path}")
         text, new_session_id = _parse_claude_output(result.stdout)
-        return AgentResult(text=read_public_response_file(response_path) or text, session_id=new_session_id)
+        return AgentResult(
+            text=read_public_response_file(response_path) or text,
+            session_id=new_session_id,
+            log_path=log_path,
+            returncode=result.returncode,
+        )
 
 
 BACKEND = ClaudeBackend()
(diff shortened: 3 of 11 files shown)


wwind123:
Implemented issue #96 and opened PR #97 against `main`.

Changes include validated marker parsing before GitHub posting, safe local failure with attempt log paths, narrow transient retry/backoff config, regression coverage, and docs updates.

Tests: `python3 -m pytest` passed with 173 tests.





-- OpenAI Codex

wwind123:
## Review: Prevent posting invalid agent diagnostics (#97)

The PR correctly addresses the issue. Here is my analysis.

### What the PR does

- Introduces `_run_validated_agent` in `orchestrator.py`, which gates all agent calls behind marker validation before any output can reach `post_pr_comment` or `post_issue_comment`.
- Adds bounded transient retry logic with configurable `--agent-max-retries` (default 2) and `--agent-retry-backoff-seconds` (default 15, 45).
- Adds `log_path` and `returncode` fields to `AgentResult` so failure context is propagated up.
- All call sites in `orchestrator.py` now use `_run_validated_agent` instead of `run_agent`, replacing the previous ad-hoc empty-check pattern.
- Adds microsecond precision to log file timestamps to prevent collision during rapid retries.
- Adds 6 regression tests covering the core failure modes.

### Correctness

**Validation gate**: `_run_validated_agent` only returns on a successful `validate(text)` call. All other paths (non-zero exit code, empty text, failed validation) either retry or raise `AgentLoopError` locally. No raw diagnostics reach GitHub in any code path.

**Transient classification**: The dual-regex approach (`TRANSIENT_AGENT_OUTPUT_RE` matched AND `NON_RETRYABLE_AGENT_OUTPUT_RE` not matched) is correct. Quota/auth/billing failures are explicitly excluded from retry. Normal prose without a required marker is not retried (validated by `test_pr_loop_does_not_retry_normal_missing_marker_response`).

**Retry bounds**: `max_attempts = config.agent_max_retries + 1` and the check `attempt < max_attempts` before sleeping ensures the loop is always bounded. The `_retry_delay` fallback to the last configured delay value for excess retries is correct.

**Session ID continuity on retry**: The original `session_id` is passed to all retry attempts unchanged, which is appropriate for stream failures where the session itself may be intact.

**Marker value threading**: `ValidatedAgentResponse.marker_value` carries the parsed result (`str` for state, `int` for PR number, `"clarification"` string) so callers no longer need to re-parse the text after validation. The `isinstance(marker_value, int)` check in `run_task_loop` is correct and cleaner than the removed `is_clarification_request` re-check.

**`run_agent` wrapper**: The old `run_agent` function in `registry.py` now raises on non-zero exit code but is no longer called from `orchestrator.py`. It is imported but unused in `cli.py` — this is pre-existing dead code, not introduced by this PR.

### Tests

All six new tests exercise the cases called out in the issue and the follow-up comment:

- Raw diagnostics not posted when all attempts fail (Gemini and plan-review paths).
- Retry on transient output, with only the valid response posted after recovery.
- Attempt log paths included in the error after exhausted retries.
- Quota failure not retried (one attempt only, no sleep commands).
- Normal prose without a marker not retried.

The `(diagnostic, 0)` vs `(diagnostic, 1)` tuple pattern in `FakeRunner._next_agent_output` correctly exercises both the zero-exit-code-with-bad-text path and the non-zero-exit-code path.

### Minor observations (not blocking)

- Empty agent output with exit code 0 is not retried because `_is_transient_agent_output("")` returns False. This is reasonable — without any diagnostic text there is no signal to classify the failure as transient.
- The `\b5\d\d\b` pattern for 5xx HTTP errors uses word boundaries correctly; it will not match 5xx codes embedded in longer identifiers like `status_500_error` but will match `503` in `HTTP 503`.

### Same-PR follow-ups

None — the implementation is clean and self-contained.

### Future follow-ups

- Consider adding retry logic (or at least non-zero exit code propagation) to `run_claude` / `run_codex` / `run_gemini` helper functions in `registry.py`, which are used for non-orchestration calls and currently bypass validation entirely.
- `run_agent` in `registry.py` is imported but never called in `cli.py`; it can be removed from that import once convenient.



-- Anthropic Claude

wwind123:
The implementation in PR #97 successfully addresses the core issue of preventing raw agent diagnostics from being posted to GitHub when an agent fails or produces a malformed response. It also introduces a robust retry mechanism for transient failures, which is a significant improvement for the reliability of the agent loop.

### Key improvements:
- **Centralized Validation:** The introduction of `_run_validated_agent` in `orchestrator.py` ensures that agent output is only posted to GitHub after it has been validated to contain the required protocol markers.
- **Transient Failure Retries:** The new retry logic with bounded backoff (defaulting to 2 retries) effectively handles common transient issues like stream interruptions, malformed tool calls, and network timeouts.
- **Improved Diagnostics:** When all retries fail, the orchestrator now provides a clear local error message that includes paths to all attempt logs, making it much easier to debug provider failures without leaking internal narration to public PR comments.
- **Microsecond Logging:** Updating the log filename timestamp to include microseconds prevents collisions when retries occur in rapid succession.
- **Comprehensive Testing:** The added tests cover various failure scenarios (transient vs. non-retryable, missing markers, empty responses) and verify that the retry logic and validation behave correctly.

The use of `runner.run(("sleep", ...))` for backoff is a clever way to keep tests fast while still verifying that delays are correctly calculated and executed.

### Same-PR follow-ups

- Remove unused imports in `src/coding_review_agent_loop/cli.py`: `run_agent`, `run_claude`, `run_codex`, and `run_gemini`.
- Remove now-unused functions in `src/coding_review_agent_loop/agents/registry.py`: `run_agent`, `run_claude`, `run_codex`, and `run_gemini` (as `orchestrator.py` and `cli.py` no longer appear to call them).


-- Google Gemini

wwind123:
Implemented and pushed the same-PR follow-ups to `fix/issue-96-safe-agent-output`.

Changes:
- Removed unused `run_agent`, `run_claude`, `run_codex`, and `run_gemini` imports from `cli.py`.
- Removed the now-unused wrapper functions from `agents/registry.py`.
- Kept `run_agent_result` intact for `orchestrator.py`.

Verification:
- `python` was unavailable in this environment.
- `python3 -m pytest` passed: 173 tests.
- `git diff --check` passed.

Commit pushed:
- `4652bbf Remove unused agent runner wrappers`

-- OpenAI Codex

wwind123:
## Review: Prevent posting invalid agent diagnostics (#97) — Round 2

The new commit (`4652bbf`) removes the dead code I flagged in round 1: `run_agent`, `run_claude`, `run_codex`, and `run_gemini` from `registry.py`, plus their corresponding unused imports from `cli.py`. The remaining `registry.py` is clean — only `get_backend`, `agent_display_name`, `agent_signature`, `default_agent_args`, and `run_agent_result` remain, all actively used.

No issues in the new commit. The round 1 analysis stands, and the same-PR follow-up concern is now resolved.



-- Anthropic Claude

wwind123:
The changes in PR #97 are well-implemented and fully address the requirements of issue #96.

By moving protocol marker validation before posting to GitHub, the orchestrator now effectively prevents internal diagnostics from leaking into public PR comments. The addition of retry logic with narrow failure classification for transient model/provider issues (like "Invalid stream" or "malformed tool call") significantly improves the reliability of the agent loop without risking infinite loops on auth or quota failures.

The new regression tests provide good coverage for the retry logic, classification, and diagnostic suppression. The inclusion of microseconds in log timestamps is a thoughtful touch to ensure log integrity during fast retries.

-- Google Gemini

PR #101, opened 2026-05-20
Force-refresh cached PR refs before reviewer runs

wwind123:
## Summary
- force-update cached `origin/pr/<number>` refs when refreshing reviewer workdirs
- keeps reviewer checkout refresh working after a PR branch is force-pushed or rewritten
- updates tests to expect the forced PR refspec

## Tests
- `/home/wwind123/tools/coding-review-agent-loop/.venv/bin/python -m pytest /home/wwind123/openai-codex/coding-review-agent-loop/tests/test_agent_loop.py -q`

## Manual verification
- `git -C /tmp/coding-review-agent-loop/wwind123-llm-dialectic/gemini/repo fetch origin +pull/144/head:refs/remotes/origin/pr/144` updated stale PR #144 ref from `34f1b4d` to `e40a5b0`.

-- OpenAI Codex

### DIFF
--- a/src/coding_review_agent_loop/config.py
+++ b/src/coding_review_agent_loop/config.py
@@ -281,10 +281,11 @@ def sync_reviewer_pr_before_review(
     default_owned = reviewer in set(config.auto_agent_dirs)
     label = f"Default {reviewer} workdir" if default_owned else option_name
     pr_ref = f"refs/remotes/origin/pr/{pr_number}"
+    pr_fetch_refspec = f"+pull/{pr_number}/head:{pr_ref}"
 
     if runner.dry_run:
         _run_git(runner, path, ("fetch", "origin"))
-        _run_git(runner, path, ("fetch", "origin", f"pull/{pr_number}/head:{pr_ref}"))
+        _run_git(runner, path, ("fetch", "origin", pr_fetch_refspec))
         _run_git(runner, path, ("checkout", "--detach", pr_ref))
         _run_git(runner, path, ("rev-parse", "HEAD"))
         _run_git(runner, path, ("status", "--short", "--branch"))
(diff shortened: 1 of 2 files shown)


wwind123:
## Review: Force-refresh cached PR refs before reviewer runs

The change is correct and well-scoped.

**What changed:** `sync_reviewer_pr_before_review` in `config.py` now uses `+pull/{n}/head:{pr_ref}` (force-update refspec) instead of the bare `pull/{n}/head:{pr_ref}`. Adding the `+` prefix tells git to update the local `refs/remotes/origin/pr/<n>` ref even when the new commit is not a fast-forward from the cached one — exactly what breaks when a PR branch is force-pushed or rebased.

**Correctness:** The fix is the standard git idiom for this. Without `+`, `git fetch origin pull/77/head:refs/remotes/origin/pr/77` will error with "non-fast-forward" if the PR was amended or rebased since the ref was last set. With `+` it unconditionally overwrites the cached ref with the current PR head, so reviewers always see the real current state.

**Variable extraction:** Defining `pr_fetch_refspec` once and reusing it in both the dry-run and live paths removes the duplication that existed before. Clean improvement.

**Tests:** All four assertions in `test_agent_loop.py` that matched the old refspec string are updated to include the `+` prefix. No stale references to the old format remain.

No blocking issues.



-- Anthropic Claude

wwind123:
## Review: Force-refresh cached PR refs before reviewer runs

The changes in this PR correctly address the issue where a reviewer's cached checkout could fail to update if a PR branch was force-pushed.

### Summary of Changes
-   Modified `sync_reviewer_pr_before_review` in `src/coding_review_agent_loop/config.py` to use a force-update refspec (`+pull/{pr_number}/head:{pr_ref}`) when fetching PR heads.
-   Refactored the code to define `pr_fetch_refspec` once, improving readability and reducing duplication.
-   Updated `tests/test_agent_loop.py` to reflect the new refspec format.

### Correctness and Impact
Using the `+` prefix in the refspec is the standard Git way to ensure that a local ref is updated even when the remote change is not a fast-forward. This is essential for a robust review loop where PR authors frequently rebase or amend their changes.

The updated tests pass successfully, confirming that the logic is correct and the expectations are met.

### Same-PR follow-ups
None.

### Future follow-ups
None.



-- Google Gemini

PR #102, opened 2026-05-20
Add pre-review local test gate

wwind123:
## Summary
- add default-on `--pre-review-tests` / `--no-pre-review-tests` behavior for `--test-command`
- run configured local tests after coder-created or coder-updated PR changes before reviewer rounds
- strengthen coder prompts to require a `Tests:` line with commands/results or an explanation
- document the new test-gate behavior

## Tests
- `/home/wwind123/tools/coding-review-agent-loop/.venv/bin/python -m pytest /home/wwind123/openai-codex/coding-review-agent-loop-pre-review-tests/tests/test_agent_loop.py -q`
- `/home/wwind123/tools/coding-review-agent-loop/.venv/bin/python -m py_compile /home/wwind123/openai-codex/coding-review-agent-loop-pre-review-tests/src/coding_review_agent_loop/*.py`

-- OpenAI Codex

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,18 @@ short-lived location. If the previous memory commit is unavailable or no longer
 diffable, the loop logs the git failure and treats all tracked files as changed
 for that refresh.
 
+Use `--test-command` to add a local test gate:
+
+```bash
+agent-loop task "Fix the flaky test" --repo OWNER/REPO --test-command "python -m pytest"
+```
+
+By default, the command runs after coder-created or coder-updated changes before
+reviewer rounds, and again after final reviewer approval before auto-merge. Add
+`--no-pre-review-tests` if you only want the final post-approval local test
+gate. The coder prompt also asks the coding agent to report the exact tests it
+ran, or explain why it could not run tests.
+
 By default Claude is the coder and Codex is the reviewer. Reverse that with:
 
 ```bash
--- a/docs/local_agent_loop.md
+++ b/docs/local_agent_loop.md
@@ -34,7 +34,8 @@ flowchart LR
     Orchestrator --> Protocol[Marker and follow-up parsing<br/>protocol.py]
     Orchestrator --> Registry[Agent registry<br/>agents/registry.py]
     Orchestrator --> GitHubOps[GitHub operations<br/>github.py]
-    Orchestrator --> OptionalTests[Optional local test command<br/>--test-command]
+    Orchestrator --> PreReviewTests[Optional pre-review local test command<br/>--test-command]
+    Orchestrator --> OptionalTests[Optional post-approval local test command<br/>--test-command]
     Orchestrator --> Followups[Approved follow-up handling<br/>summaries / issues / same-PR fixes]
 
     Registry --> Claude[Claude backend<br/>claude]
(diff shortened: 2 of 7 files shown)


wwind123:
## Review: Add pre-review local test gate

This is a clean, well-structured addition. The feature rationale is sound — running the local test gate before reviewer rounds avoids wasting review cycles on already-failing code.

**Implementation correctness**

- `run_pre_review_tests` correctly double-guards on `config.pre_review_tests` and `config.test_command`, so it is a no-op when either condition is absent.
- The `pre_review_test_pending` flag propagates correctly through all three entry paths: `run_issue_loop`, `run_task_loop` (via `_run_plan_first_loop`), and the coder follow-up inside `run_pr_loop`. The `pr` subcommand calling `run_pr_loop` directly leaves `pre_review_test_pending=False`, which is the right default — when you point the tool at an existing PR, there is no guarantee the coder "just pushed."
- The `--pre-review-tests` / `--no-pre-review-tests` mutually exclusive group uses `default=True` on the `store_true` action, which is the correct argparse pattern for a default-on boolean flag.

**Tests**

Three new tests cover:
1. Pre-review tests fire in the right sequence and the right number of times (`first_coder < first_test < first_review`, count == 3).
2. `--no-pre-review-tests` disables the pre-review gate while leaving the post-approval gate intact.
3. CLI defaults wire through `config_from_args` correctly.

The updated `test_pr_loop_routes_migration_validation_failure_through_coder_followup` assertion (`first_review_index < pytest_index < second_review_index`) correctly reflects the new ordering: the first pytest invocation is now the pre-review gate before round 2 (after the coder pushes the migration fix), and `commands.index` / `command_index` find that first occurrence.

**One minor note (not blocking)**

`_coder_test_reporting_guidance()` is injected into `build_task_clarification_prompt`. During a clarification round the coder is not yet implementing, so the "run tests before opening the PR" guidance reads slightly out of place. In practice the coder will simply skip it if there is nothing to test, so this is harmless.

CI status: **passing**.



-- Anthropic Claude

wwind123:
Warning: True color (24-bit) support not detected. Using a terminal with true color enabled will result in a better visual experience.
YOLO mode is enabled. All tool calls will be automatically approved.
YOLO mode is enabled. All tool calls will be automatically approved.
Ripgrep is not available. Falling back to GrepTool.
(node:142579) [DEP0190] DeprecationWarning: Passing args to a child process with shell option true can lead to security vulnerabilities, as the arguments are not escaped, only concatenated.
(Use `node --trace-deprecation ...` to show where the warning was created)
PR #102 introduces a "pre-review local test gate" to the agent loop, allowing users to configure a local test command that runs automatically after coder changes but before reviewer rounds. This is a great efficiency improvement as it prevents reviewers (and the user's budget) from being spent on code that already fails the local test suite.

Key observations:
- **CLI Flags**: Added `--pre-review-tests` (default) and `--no-pre-review-tests` to control the behavior of `--test-command`.
- **Orchestration**: Correctly manages the `pre_review_test_pending` state across various loops (`issue`, `task`, `plan-first`, and `pr`).
- **Prompts**: Updated coder prompts to include guidance on local test reporting (`Tests:` line), which encourages agents to be proactive.
- **Robustness**: The gate correctly raises an `AgentLoopError` if tests fail, halting the loop before reviewers are invoked.
- **Coverage**: Added several tests in `tests/test_agent_loop.py` verifying the execution order and the ability to disable the pre-review gate.

The implementation is surgical, follows existing patterns, and correctly addresses the goal.

### Same-PR follow-ups
None.

### Future follow-ups
None.

-- Google Gemini

PR #110, opened 2026-05-21
Validate Alembic migration topology before PR approval

### LINKED ISSUE #109: Validate migration head topology when PR adds Alembic revisions
## Bug / hardening gap

The agent loop can approve a PR that adds a new Alembic migration with the wrong `down_revision`, creating multiple migration heads even though the Git branch itself is current and linear.

## Concrete example

In `wwind123/llm-dialectic` PR #156:

- the PR branch ancestry was correct and based on then-current `main`
- Claude added `alembic/versions/e4f5a6b7c8d9_add_gemini_3_5_flash_pricing.py` during review follow-up
- that file declared `down_revision = "5d5f0e1a2b3c"`
- but `main` already had newer Alembic revisions on top of that chain: `a6b7c8d9e0f1` and then `402b9e8af79b`

So PR #156 created two Alembic heads:

- `402b9e8af79b`
- `e4f5a6b7c8d9`

PR #157 was then needed to add a merge migration so `alembic upgrade head` would work again.

## Why this matters

This is easy for the coder agent to get wrong when adding a migration in a follow-up round, especially if it copies the parent revision from an older nearby file instead of checking the current repo head.

The current review flow may miss it because:

- Git branch freshness can still be correct
- code reviewers may focus on application behavior and not inspect Alembic topology
- no explicit migration-head validation is run before approval/merge

This can break deploys with a multiple-heads error.

## Expected behavior

When a PR adds or modifies migration files under `alembic/versions/`, the orchestrator should detect migration topology problems before the PR is treated as approved or merge-ready.

## Possible approach

1. Add migration-aware validation when changed files include `alembic/versions/*.py`.
2. Run project-specific commands if available, for example:
   - `alembic heads`
   - `alembic upgrade head`
3. Fail or bounce back to the coder if:
   - more than one head exists unexpectedly
   - the new migration does not descend from the current head when a single linear chain is expected
   - upgrade-to-head fails
4. Surface the exact migration ancestry mismatch in the orchestrator log/comment so the coder can fix `down_revision` or add an intentional merge migration.

## Scope notes

This should probably be configurable or heuristic-based, since not every repo uses Alembic and some repos may intentionally allow multiple heads. But for a repo with an established single-head migration chain, this should be treated as a blocking validation error.

## Acceptance criteria

- Reproduce a PR like `llm-dialectic` PR #156 in tests.
- Detect when a newly added Alembic revision branches from an older ancestor instead of the current head.
- Detect multiple-head failures before approval/merge.
- Log a clear actionable message pointing at the offending migration and expected current head.
- Avoid affecting repos that do not use Alembic.

-- OpenAI Codex

wwind123:
## Summary
- add Alembic migration topology validation that runs only when a PR changes \
- refresh the active coder checkout to the PR head before running post-approval validation
- cover stale \, malformed metadata, linear success, merge-head skip, and orchestrator follow-up behavior in tests

Closes #109

### DIFF
--- a/src/coding_review_agent_loop/config.py
+++ b/src/coding_review_agent_loop/config.py
@@ -280,6 +280,50 @@ def sync_reviewer_pr_before_review(
     option_name = _agent_dir_option(reviewer)
     default_owned = reviewer in set(config.auto_agent_dirs)
     label = f"Default {reviewer} workdir" if default_owned else option_name
+    sync_checkout_to_pr(
+        config,
+        runner,
+        path=path,
+        label=label,
+        default_owned=default_owned,
+        pr_number=pr_number,
+        pr_metadata=pr_metadata,
+    )
+
+
+def sync_coder_pr_before_validation(
+    config: AgentLoopConfig,
+    runner: Runner,
+    pr_number: int,
+    pr_metadata: PullRequestMetadata,
+) -> None:
+    """Refresh the active coder checkout to the current PR head before local validation."""
+    path = active_workdir(config)
+    option_name = _agent_dir_option(config.coder)
+    default_owned = config.coder in set(config.auto_agent_dirs)
+    label = f"Default {config.coder} workdir" if default_owned else option_name
+    sync_checkout_to_pr(
+        config,
+        runner,
+        path=path,
+        label=label,
+        default_owned=default_owned,
+        pr_number=pr_number,
+        pr_metadata=pr_metadata,
+    )
+
+
+def sync_checkout_to_pr(
+    config: AgentLoopConfig,
+    runner: Runner,
+    *,
+    path: Path,
+    label: str,
+    default_owned: bool,
+    pr_number: int,
+    pr_metadata: PullRequestMetadata,
+) -> None:
+    """Refresh a checkout to the current PR head."""
     pr_ref = f"refs/remotes/origin/pr/{pr_number}"
     pr_fetch_refspec = f"+pull/{pr_number}/head:{pr_ref}"
 
--- a/src/coding_review_agent_loop/migrations.py
+++ b/src/coding_review_agent_loop/migrations.py
@@ -1,0 +1,267 @@
+"""Alembic migration topology validation helpers."""
+
+from __future__ import annotations
+
+import ast
+from dataclasses import dataclass
+from pathlib import Path
+
+from .config import AgentLoopConfig, _run_git
+from .errors import AgentLoopError
+from .github import PullRequestMetadata
+from .runner import Runner
+
+ALEMBIC_VERSIONS_PREFIX = "alembic/versions/"
+
+
+@dataclass(frozen=True)
+class RevisionMetadata:
+    path: str
+    revision: str
+    down_revisions: tuple[str, ...]
+
+
+@dataclass(frozen=True)
+class RevisionIndex:
+    by_revision: dict[str, RevisionMetadata]
+    by_path: dict[str, RevisionMetadata]
+
+
+@dataclass(frozen=True)
+class MigrationValidationResult:
+    ok: bool
+    message: str | None = None
+
+
+def validate_pr_migration_topology(
+    runner: Runner,
+    *,
+    config: AgentLoopConfig,
+    checkout: Path,
+    pr_metadata: PullRequestMetadata,
+) -> MigrationValidationResult:
+    base_branch = (pr_metadata.base_branch or config.base).strip() or config.base
+    base_ref = f"origin/{base_branch}"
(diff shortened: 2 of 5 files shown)


wwind123:
Implemented on branch `codex/issue-109-alembic-validation` and opened [PR #110](https://github.com/wwind123/coding-review-agent-loop/pull/110).

The change adds a post-approval Alembic topology gate that only runs when the PR diff touches `alembic/versions/*.py`, refreshes the active coder checkout to the PR head before validation, and blocks approval with an actionable message when a new revision branches from an older ancestor or changed migration metadata cannot be validated. Reviewer guidance was also tightened, and the test suite now covers the stale-`down_revision` regression, linear success, pre-existing multi-head skip behavior, malformed metadata, and the orchestrator follow-up loop.

Tests run: `python3 -m pytest`




-- OpenAI Codex

wwind123:
## Review of PR #110 — Validate Alembic migration topology before PR approval

The implementation is solid and correctly addresses the issue. The AST-based static analysis approach avoids needing Alembic installed, the topology logic is correct, and the test suite covers the exact regression from the issue (`down_revision` pointing at a stale ancestor), linear success, pre-existing multi-head skip, malformed metadata, and the orchestrator round-trip. No blocking issues.

### Same-PR follow-ups

**1. `is_alembic_version_path` admits `__init__.py` and other non-migration files**

`migrations.py:99–100` — the filter passes any `alembic/versions/**/*.py`. If a repo has `alembic/versions/__init__.py` (empty or containing imports), `parse_revision_metadata` raises `AgentLoopError: missing 'revision'`, which propagates as a false-positive `ok=False` result. The error message would say something like "missing `revision` in `alembic/versions/__init__.py`" — confusing and blocking for a PR that doesn't actually create a migration topology problem.

Standard Alembic migration filenames follow the convention `{hex_revision}_{slug}.py`. Tightening the filter, or gracefully skipping files that lack both `revision` and `down_revision` assignments (i.e. not counting them as migration files rather than erroring on them), would prevent this class of false positive. The simplest targeted fix is to exclude `__init__.py` and `env.py` explicitly, or require the file stem to match a hex-prefix pattern.

**2. `RevisionIndex` is declared `frozen=True` but its fields are mutable dicts**

`migrations.py:24–27` — `frozen=True` on the dataclass prevents attribute reassignment but does nothing to prevent mutation of the `by_revision` and `by_path` dicts themselves. The `frozen=True` annotation is therefore misleading. Either drop it (since the class is only used internally and never mutated in practice) or replace the dict fields with `types.MappingProxyType` to back the immutability claim.

### Future follow-ups

**1. Complement static analysis with `alembic heads` when available**

The issue's acceptance criteria mention running `alembic heads` as an alternative detection strategy. The current AST approach will miss topology problems that involve non-literal revision references (which it already blocks correctly) but also scenarios where the alembic directory structure doesn't match the `alembic/versions/` convention, or where the installed Alembic disagrees with the static graph. A configurable probe — try `alembic heads`; fall back to AST if not available — would make the gate more authoritative and closer to what the failing deployment would actually see.

**2. Allow explicit validation opt-out in config**

The check is currently gated purely on whether `alembic/versions/*.py` files appear in the diff, with a pass-through when the base already has multiple heads. Some repos intentionally maintain multiple named branches. A boolean flag (e.g. `validate_migration_topology: bool = True`) would let those repos opt out cleanly without needing to work around the heuristic.

-- Anthropic Claude

wwind123:
The implementation of Alembic migration topology validation is robust, well-integrated, and directly addresses the issue of accidental multi-head migrations. By utilizing AST parsing to inspect migration files, the agent loop can now detect branching errors before they reach production.

Key strengths of this PR:
- **Surgical Integration:** The validation is inserted into the orchestrator's loop at the right moment—after code reviews are complete but before final approval.
- **Clear Error Messaging:** When a validation failure occurs, the orchestrator provides a detailed message identifying the offending revision and the expected current head, which is then routed back to the coder agent for correction.
- **Robust Heuristics:** The tool correctly handles cases where the base branch might already have multiple heads or where the PR is adding the first migrations to a repository.
- **Thorough Testing:** The new tests cover success scenarios, intentional branching/merging, and various failure modes (wrong parent, non-literal metadata, etc.).

I have verified the changes by running the full test suite, and all 191 tests passed successfully.

### Future follow-ups

- Consider supporting multiple Alembic database "engines" or "versions" directories if the project structure uses non-standard paths.
- Add similar topology validation for other migration frameworks (e.g., Django, Flyway) if they become commonly used in this agent loop's target repositories.



-- Google Gemini

wwind123:
Created approved-review future follow-up issues for PR #110:

- https://github.com/wwind123/coding-review-agent-loop/issues/111
- https://github.com/wwind123/coding-review-agent-loop/issues/112

-- coding-review-agent-loop

PR #122, opened 2026-05-22
Tolerate stray future followups in blocking reviews

wwind123:
## Summary
- ignore stray Future follow-ups sections when a review is otherwise blocking
- tighten the reviewer prompt so blocking reviews explicitly omit future follow-ups
- add parser and PR-loop regressions for the Gemini failure mode

## Testing
- /home/wwind123/tools/coding-review-agent-loop/.venv/bin/python -m pytest /home/wwind123/openai-codex/coding-review-agent-loop-pre-review-tests/tests/test_agent_loop.py -q
- /home/wwind123/tools/coding-review-agent-loop/.venv/bin/python -m py_compile /home/wwind123/openai-codex/coding-review-agent-loop-pre-review-tests/src/coding_review_agent_loop/*.py

-- OpenAI Codex

### DIFF
--- a/src/coding_review_agent_loop/prompts.py
+++ b/src/coding_review_agent_loop/prompts.py
@@ -643,7 +643,9 @@ heading:
 
 Same-PR follow-ups will be sent back to {coder_name} and require another review
 round before final approval. Do not put trivial style nits in either follow-up
-section.
+section. If you return `<!-- AGENT_STATE: blocking -->`, do not include a
+Future follow-ups section at all; keep only blocking findings and any Same-PR
+follow-ups that should be fixed before approval.
 """
     else:
         followup_guidance = """If you approve but notice substantial work that is better handled separately in
--- a/src/coding_review_agent_loop/prompts.py
+++ b/src/coding_review_agent_loop/prompts.py
@@ -643,9 +643,9 @@ heading:
 
 Same-PR follow-ups will be sent back to {coder_name} and require another review
 round before final approval. Do not put trivial style nits in either follow-up
-section. If you return `<!-- AGENT_STATE: blocking -->`, do not include a
-Future follow-ups section at all; keep only blocking findings and any Same-PR
-follow-ups that should be fixed before approval.
+section. If you return `<!-- AGENT_STATE: blocking -->`, do not use structured
+follow-up sections; include all findings in the main body of your response so
+they are not missed during revision.
 """
     else:
         followup_guidance = """If you approve but notice substantial work that is better handled separately in
--- a/src/coding_review_agent_loop/protocol.py
+++ b/src/coding_review_agent_loop/protocol.py
@@ -258,7 +258,10 @@ def parse_review(text: str, *, reviewer: str) -> ParsedReview:
     followups = parse_approved_followups(text, reviewer=reviewer)
     dispositions = parse_unresolved_item_dispositions(text, reviewer=reviewer)
     if state == "blocking" and followups.future:
-        raise AgentLoopError("Blocking reviews may not include Future follow-ups.")
+        # Some agents still append speculative future work to otherwise valid
+        # blocking reviews. Preserve the blocking findings and ignore that
+        # lower-priority section rather than aborting the whole round.
+        followups = ApprovedFollowups(same_pr=followups.same_pr, future=())
     if state == "blocking" and any(item.disposition == "future" for item in dispositions):
         raise AgentLoopError(
             "Blocking reviews may not downgrade prior unresolved items to Future follow-ups."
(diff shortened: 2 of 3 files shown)


wwind123:
## Review: Tolerate stray future followups in blocking reviews

The change is correct and the fix is well-scoped. Here's my analysis of each part:

### `protocol.py` — core behavioral change

`parse_review` now silently drops `followups.future` when `state == "blocking"` instead of raising `AgentLoopError`. This is safe because the orchestrator already cannot reach the code that consumes `parsed_review.followups.future` for blocking reviews — the `continue` at orchestrator.py:1265 exits the per-reviewer loop before line 1269 where future followups are processed. The former hard error was protecting an invariant that didn't actually need protecting. Same-pr followups in `followups.same_pr` are preserved, which matters because they represent actionable pre-merge cleanup even in a blocking context.

The one remaining hard error — `disposition == "future"` for prior unresolved items — is correctly left in place; that's a different and meaningful constraint.

### `prompts.py` — preventive guidance

The new sentence is placed correctly in the `fix-and-*` branch (where both `### Same-PR follow-ups` and `### Future follow-ups` are described). The instruction is explicit about which section to omit and what to keep, so reviewers receiving this prompt have a clear rule to follow going forward.

### `tests/test_agent_loop.py` — coverage

- **Unit test** (`test_parse_review_drops_future_followups_in_blocking_reviews`): Renamed accurately, adds a `### Same-PR follow-ups` section to confirm preservation, asserts `future == ()`. Solid.
- **Prompt assertion**: The added assert in `test_review_prompt_allows_same_pr_followups_for_fix_modes` verifies the new instruction verbatim.
- **Integration test** (`test_pr_loop_keeps_blocking_review_when_future_followups_are_misclassified`): Covers the full PR loop. The two assertions that matter most are that the loop returns 0 (wasn't aborted) and that `"Consider a broader cleanup later."` appears in the posted comment (raw reviewer text is preserved as-is for humans). The followup-prompt assertion confirms the blocking content reaches the coder.

CI is passing. No blocking issues.



-- Anthropic Claude

wwind123:
The changes correctly address the issue where agents mistakenly include "Future follow-ups" in blocking reviews, which previously caused the entire round to fail with an `AgentLoopError`. By tolerating and dropping these lower-priority items in the parser, the loop can continue with the primary blocking findings. The added tests verify both the parser behavior and the end-to-end PR loop integration.

I have one minor suggestion regarding the prompt guidance to ensure agents don't rely on the "Same-PR follow-ups" section for blocking findings, as those items are currently stripped from the structured prompt sent back to the coder in the next round.

### Same-PR follow-ups

- In `src/coding_review_agent_loop/prompts.py`, the new guidance says: "keep only blocking findings and any Same-PR follow-ups that should be fixed before approval." However, the orchestrator's `_review_freeform_summary_text` strips the `### Same-PR follow-ups` section when creating the blocking summary. This means any items placed there by a blocking reviewer will be posted to GitHub but won't appear in the structured prompt sent to the coder in the next round. Consider updating the guidance to: "If you return `<!-- AGENT_STATE: blocking -->`, do not use structured follow-up sections; include all findings in the main body of your response so they are not missed during revision."

-- Google Gemini

wwind123:
Updated PR #122 on `codex/fix-blocking-future-followups` and pushed commit `2c39a33`.

The change stays narrow: `build_review_prompt()` now tells reviewers that if they return `<!-- AGENT_STATE: blocking -->`, they must not use structured follow-up sections and should put all blocking findings in the main body so the coder round-trip cannot drop them. I also updated the existing fix-mode prompt test to assert the new wording.

Tests: `python3 -m pytest tests/test_agent_loop.py -k 'review_prompt_allows_same_pr_followups_for_fix_modes or review_freeform_summary_text_strips_structured_followup_sections'` passed; `python3 -m pytest tests/test_agent_loop.py` passed.

-- OpenAI Codex

wwind123:
## Review: Tolerate stray future followups in blocking reviews (round 2)

- [item-1] resolved

The prompt guidance in `prompts.py` now reads exactly as Gemini suggested: "If you return `<!-- AGENT_STATE: blocking -->`, do not use structured follow-up sections; include all findings in the main body of your response so they are not missed during revision." This correctly steers blocking reviewers away from `### Same-PR follow-ups` (which `_review_freeform_summary_text` strips) and into the freeform body (which is preserved in the coder prompt).

The two new test assertions in `test_review_prompt_allows_same_pr_followups_for_fix_modes` match the updated wording verbatim. No other changes in this round.

**Note on CI**: GitHub checks are pending (test queued) as of this review. I cannot confirm tests are globally passing, but the round 2 diff is limited to the prompt string and the matching test assertions — there is no logic change relative to round 1 where tests passed.



-- Anthropic Claude

PR #159, opened 2026-05-24
Fix same-round reviewer item leakage

### LINKED ISSUE #158: Fix same-round item leakage between reviewers in multi-reviewer rounds
## Summary

In multi-reviewer rounds, a later reviewer can see newly created `item-N` records from an earlier reviewer in the same round and then disposition those items as if they were prior carried-forward items. The validator rejects them as unknown IDs, causing the run to fail.

## Concrete example

Issue #152 (plan-first flow):

- Round 1 created `item-1` through `item-9`.
- In round 2, Gemini reviewed first and:
  - marked `item-1` through `item-9` as resolved
  - created new round-2 items `item-14` through `item-17`
- Then Claude reviewed in the same round and returned dispositions for `item-14` through `item-17`.
- The orchestrator rejected this with:
  - `Plan review referenced unknown prior unresolved plan item IDs: item-14, item-15, item-16, item-17`

## Why this happens

The orchestrator defines the "prior unresolved item" set for a reviewer as the ledger at the start of the round. Newly created items from earlier reviewers in that same round are not part of that carried-forward set.

However, later reviewers can still see earlier same-round comments in the GitHub issue/PR thread, including the newly assigned `item-N` IDs. That makes it natural for a later reviewer to disposition them, even though the validator does not allow it.

## Problem

This creates a protocol mismatch:

- reviewer-visible context includes same-round new items from earlier reviewers
- validator only accepts carried-forward items that existed before the round began

So later reviewers can fail the run just by reacting to visible current-round context.

## Expected behavior

Agent-loop should choose one clear model and enforce it consistently:

### Option A: isolate same-round items
- Later reviewers should not be asked to disposition newly created items from earlier reviewers in the same round.
- The prompt/context shown to later reviewers should avoid presenting those item IDs as if they are part of the carried-forward ledger.

### Option B: allow same-round shared item reconciliation
- If later reviewers are allowed to see and disposition same-round items from earlier reviewers, then the orchestrator/validator must explicitly support that and merge those items into the active round ledger before validating later reviewers.

Either approach is acceptable, but the current mixed behavior is brittle.

## Why this matters

This is a recurring failure mode for multi-reviewer rounds:

- later reviewers act on visible same-round context
- validator rejects that context as unknown
- run aborts

It also makes resumability and round-state reconstruction harder, because "what counts as the active ledger" is no longer consistent across reviewers in the same round.

-- OpenAI Codex

wwind123:
## Summary
- stop appending same-round tracked `item-N` sections to public reviewer comments
- tell plan and PR reviewers that only the carried-forward prior-unresolved section is dispositionable
- add regressions for later reviewers in the same round plus updated comment-shape assertions

## Testing
- `python3 -m pytest tests/test_agent_loop.py`

Fixes #158

### DIFF
--- a/src/coding_review_agent_loop/orchestrator.py
+++ b/src/coding_review_agent_loop/orchestrator.py
@@ -1001,12 +1001,6 @@ def _render_public_review_comment(
                 dispositions=dispositions,
             ),
         )
-    if new_items:
-        section_lines = ["### New tracked unresolved items"]
-        section_lines.extend(
-            f"- [{item.item_id}] {_format_unresolved_item_label(item)}" for item in new_items
-        )
-        rendered = _append_before_trailing_metadata(rendered, "\n".join(section_lines))
     return rendered
 
 
--- a/src/coding_review_agent_loop/prompts.py
+++ b/src/coding_review_agent_loop/prompts.py
@@ -545,6 +545,11 @@ active for this planning round. If you return `<!-- AGENT_PLAN_STATE:
 blocking -->`, do not use structured Future follow-ups; keep all required
 current-round issues in the main blocking content so they are not missed
 during plan revision.
+Only items listed under `Prior unresolved plan items from earlier rounds` are
+eligible for dispositions in this round. If same-round findings from other
+reviewers appear elsewhere in the issue discussion, treat them as
+informational only and do not disposition them until a later round carries
+them forward explicitly.
 {unresolved_items_guidance}
 Use blocking only when the current plan still has blocking plan issues or
 same-plan follow-ups. All configured reviewers ({reviewer_group}) must approve
@@ -867,6 +872,11 @@ When the PR changes files under `alembic/versions/`, verify migration topology:
 new revisions should descend from the current head unless the PR intentionally
 adds a merge migration.
 {human_requirements_guidance}
+Only items listed under `Prior unresolved review items from earlier rounds`
+are eligible for dispositions in this round. If same-round findings from
+other reviewers appear elsewhere in the PR discussion, treat them as
+informational only and do not disposition them until a later round carries
+them forward explicitly.
 {unresolved_items_guidance}
 {followup_guidance}
 Use blocking only for issues that should prevent merge.
--- a/tests/test_agent_loop.py
+++ b/tests/test_agent_loop.py
@@ -1888,6 +1888,8 @@ def test_review_prompt_includes_prior_unresolved_items_and_disposition_instructi
     assert "- [item-id] resolved" in prompt
     assert "Only use `future follow-up` when returning `approved`." in prompt
     assert "Contradictory forms like `same-pr: none`, `still blocking: none`, and `future follow-up: none` are invalid" in prompt
+    assert "Only items listed under `Prior unresolved review items from earlier rounds`" in prompt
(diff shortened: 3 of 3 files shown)


wwind123:
Implemented the `#158` fix on branch `fix-158-same-round-item-isolation` and opened PR #159 against `main`.

The change takes Option A: public reviewer comments no longer expose same-round tracked `item-N` IDs, while hidden `new_items` metadata remains intact for resume and next-round carry-forward. I also tightened both plan-review and PR-review prompts so reviewers only disposition items from the explicit prior-unresolved section, and added regressions for both multi-reviewer plan rounds and PR rounds.

Tests: `python3 -m pytest tests/test_agent_loop.py` — passed



-- OpenAI Codex

wwind123:
The PR correctly implements isolation for reviewer items within the same round. By removing the public "New tracked unresolved items" section and updating the reviewer prompts, it prevents reviewers from attempting to disposition findings from their peers in the same round before those findings are officially tracked as prior items.

### Highlights:
- **Isolation**: New findings no longer append item IDs to public comments during the round they are first reported.
- **Prompt Guardrails**: AI reviewers are now explicitly instructed to only disposition items from earlier rounds.
- **Persistence**: Items are still correctly tracked via hidden `AGENT_LOOP_META` comments, ensuring they are available for disposition in subsequent rounds.
- **Test Coverage**: New tests verify that item IDs from an earlier reviewer in the same round are not exposed to subsequent reviewers.

All tests passed.

-- Google Gemini

wwind123:
## Review: Fix same-round reviewer item leakage

**CI:** passing ✅ | **Local tests:** 319 passed ✅

The fix is correct, minimal, and well-tested.

### What the change does

**`orchestrator.py`** — Removes the `"### New tracked unresolved items"` block from `_render_public_review_comment`. The `new_items` are still serialised into the hidden `AGENT_LOOP_META` base-64 payload (via `PostedRoundMetadata.new_items`), so item tracking and round carry-forward are fully preserved. Only the public section that would have broadcast same-round item IDs to subsequent reviewers is gone.

**`prompts.py`** — Adds an explicit guardrail paragraph to both `build_plan_review_prompt` and `build_review_prompt` telling AI reviewers that only items under the `Prior unresolved … items from earlier rounds` heading are eligible for disposition in the current round. Same-round findings are informational until the next round carries them forward.

**`tests/test_agent_loop.py`** — Updates three existing tests to match the new expected comment shape (no `"### New tracked unresolved items"` section), and adds two targeted regression tests:
- `test_pr_loop_does_not_expose_same_round_item_ids_to_later_reviewers` — verifies `[item-1]` absent from the second reviewer's prompt and from the first reviewer's posted comment.
- `test_issue_loop_plan_first_does_not_expose_same_round_item_ids_to_later_reviewers` — same guarantee for the plan-review path.

### Minor observation (non-blocking)

`_render_public_review_comment` still declares `new_items: Sequence[UnresolvedReviewItem]` in its signature but the body no longer reads it. The parameter is harmless—all call sites still supply it correctly and the hidden metadata path still uses the value—but it is now dead from the function's perspective. A future cleanup pass could drop it from the signature and call sites.

This does not affect correctness or security, so I am not marking it blocking.

### Summary

The fix cleanly solves the isolation problem without disrupting resume, carry-forward, or any other round-management logic. All 319 tests pass.



-- Anthropic Claude

PR #182, opened 2026-05-30
Use Gemini CLI for repair pass instead of google-genai SDK

### LINKED ISSUE #180: Repair pass should use the Gemini CLI instead of the google-genai SDK
## Problem

The repair pass introduced in PR #176 calls `gemini-3.1-flash-lite` via the `google-genai` Python SDK, which requires `GEMINI_API_KEY` to be set in the environment. When the key is absent the repair silently returns `None` and the original format error is surfaced as if no repair was attempted — with no warning to the user.

This was observed in practice: running `agent-loop pr 179` triggered the repair pass ("attempting repair pass" logged) but `GEMINI_API_KEY` was not in the agent-loop environment, so `attempt_repair` returned `None` silently, the repair was skipped, and Gemini's malformed response was not recovered.

## Root cause

When Gemini is used as a **reviewer**, agent-loop invokes the **Gemini CLI** (`gemini --prompt ...`) as a subprocess. The CLI uses its own Google account authentication (OAuth, set up at install time) — no `GEMINI_API_KEY` needed.

The repair pass uses a completely different auth path (the `google-genai` SDK), creating an unnecessary and surprising dependency that silently fails.

## Fix

Have `attempt_repair` invoke the Gemini CLI the same way the reviewer does, using the `-m/--model` flag to select `gemini-3.1-flash-lite`:

```python
import subprocess
from .agents.gemini import _strip_gemini_preamble

def attempt_repair(raw: str, gemini_cmd: str) -> str | None:
    prompt = _REPAIR_PROMPT.replace("{raw_response}", raw, 1)
    try:
        result = subprocess.run(
            [gemini_cmd, "--model", "gemini-3.1-flash-lite", "--prompt", prompt],
            capture_output=True, text=True, timeout=120,
        )
    except Exception:
        return None
    text = _strip_gemini_preamble(result.stdout.strip())
    return text or None
```

The orchestrator already has `config.gemini_cmd` available at every Gemini call site, so passing it through to `attempt_repair` is straightforward.

## Benefits

- **Same auth as the reviewer** — works wherever `agent-loop` works, no extra setup
- **No extra dependency** — the `google-genai` SDK requirement can be removed from `pyproject.toml`
- **Explicit lighter model** — `--model gemini-3.1-flash-lite` is specified at call time
- **No worktree needed** — the repair prompt is pure text-in/text-out with no tool calls, so `--yolo`/`--skip-trust` are not required
- **Fails visibly** — a non-zero returncode or empty output is easy to detect and log

## Affected PRs

- PR #176 (`repair.py` introduced) — uses SDK, silently fails without `GEMINI_API_KEY`
- PR #179 (extends repair to coder follow-ups) — inherits the same problem

This issue should be addressed before or alongside PR #179 to ensure the repair pass actually works in the standard agent-loop environment.

-- Anthropic Claude

wwind123:
## Summary

- Replaces the `google-genai` SDK call in `attempt_repair` with a `subprocess.run` invocation of the Gemini CLI, using `--model gemini-3.1-flash-lite`
- Passes `config.gemini_cmd` from the orchestrator through to `attempt_repair` so the same binary is used as the reviewer
- Removes the `google-genai>=1.0` dependency from `pyproject.toml`
- Updates tests to verify the CLI path instead of the SDK mock

## Why

The old SDK approach required `GEMINI_API_KEY` to be set. When it was absent, `attempt_repair` silently returned `None` and the format error was surfaced as-is — no warning, no recovery. The Gemini CLI uses OAuth credentials set up at install time, the same auth already working for review runs.

Fixes #180

## Test plan

- [ ] `test_attempt_repair_returns_none_when_subprocess_fails` — subprocess non-zero exit → `None`
- [ ] `test_attempt_repair_returns_none_when_subprocess_raises` — e.g. binary not found → `None`
- [ ] `test_attempt_repair_calls_cli_and_returns_text` — verifies `gemini --model gemini-3.1-flash-lite --prompt <prompt>` is called and output is returned
- [ ] All existing orchestrator repair integration tests updated for new signature and passing
- [ ] Full test suite: 427 passed

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,9 +20,7 @@ classifiers = [
   "Programming Language :: Python :: 3.12",
 ]
 
-dependencies = [
-  "google-genai>=1.0",
-]
+dependencies = []
 
 [project.optional-dependencies]
 dev = ["pytest>=8"]
--- a/src/coding_review_agent_loop/agents/gemini.py
+++ b/src/coding_review_agent_loop/agents/gemini.py
@@ -14,7 +14,7 @@ from .base import (
     with_public_response_file_instruction,
 )
 from ..logging import agent_log_path, log
-from ..protocol import CLARIFY_RE, STATE_RE
+from ..protocol import CLARIFY_RE, PLAN_STATE_RE, STATE_RE
 from ..runner import Runner
 from ..usage import UsageMetadata, coerce_int, first_present
 
@@ -55,7 +55,7 @@ def _strip_gemini_preamble(raw: str) -> str:
     if marker_stripped != raw:
         return marker_stripped
 
-    marker_matches = [*STATE_RE.finditer(raw), *CLARIFY_RE.finditer(raw)]
+    marker_matches = [*STATE_RE.finditer(raw), *PLAN_STATE_RE.finditer(raw), *CLARIFY_RE.finditer(raw)]
     if not marker_matches:
         return raw
 
--- a/src/coding_review_agent_loop/orchestrator.py
+++ b/src/coding_review_agent_loop/orchestrator.py
@@ -329,7 +329,7 @@ def _run_validated_agent(
                 )
                 if use_repair and not _is_transient_agent_output(text):
                     log(config, f"{agent_name}: schema validation failed ({exc}); attempting repair pass")
-                    repaired = attempt_repair(text)
+                    repaired = attempt_repair(text, config.gemini_cmd)
                     if repaired is not None:
                         try:
                             marker_value = validate(repaired)
--- a/src/coding_review_agent_loop/repair.py
+++ b/src/coding_review_agent_loop/repair.py
@@ -11,7 +11,7 @@ from __future__ import annotations
 import logging
 import subprocess
 
-from .agents.gemini import _strip_gemini_preamble
+from .agents.gemini import _parse_gemini_payload
 
 _logger = logging.getLogger(__name__)
 
@@ -139,5 +139,5 @@ def attempt_repair(raw: str, gemini_cmd: str) -> str | None:
     if result.returncode != 0:
         _logger.debug("repair pass CLI exited with code %d", result.returncode)
         return None
-    text = _strip_gemini_preamble(result.stdout.strip())
+    text, _, _, _ = _parse_gemini_payload(result.stdout.strip())
     return text or None
--- a/src/coding_review_agent_loop/repair.py
+++ b/src/coding_review_agent_loop/repair.py
@@ -9,7 +9,9 @@ as blocking per the issue guardrails.
 from __future__ import annotations
 
 import logging
-import os
+import subprocess
+
+from .agents.gemini import _strip_gemini_preamble
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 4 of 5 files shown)


wwind123:
PR #182 fixes this by replacing the `google-genai` SDK call in `attempt_repair` with a `subprocess.run` invocation of the Gemini CLI.

**Changes:**

- `repair.py`: drops `os`/`google-genai` imports, adds `subprocess` + `_strip_gemini_preamble`; new signature `attempt_repair(raw, gemini_cmd)` runs `[gemini_cmd, "--model", "gemini-3.1-flash-lite", "--prompt", prompt]`; returns `None` on non-zero exit or empty output
- `orchestrator.py`: passes `config.gemini_cmd` to `attempt_repair`
- `pyproject.toml`: removes `google-genai>=1.0` dependency
- `tests/test_agent_loop.py`: replaces SDK mock tests with subprocess mock tests; updates `fake_attempt_repair` signatures in orchestrator integration tests

All 427 tests pass.

Tests: `python3 -m pytest tests/test_agent_loop.py -q` → 427 passed

-- Anthropic Claude

wwind123:
**Review verdict:** Blocking

```json
{
  "schema_version": 1,
  "kind": "pr_review",
  "state": "blocking",
  "summary": "The PR successfully migrates the repair pass from the google-genai SDK to the Gemini CLI, which unifies authentication and removes the `google-genai` dependency. However, the new implementation is less robust than the standard Gemini backend regarding JSON output and plan review markers.",
  "blocking_items": [
    "The `attempt_repair` function in `src/coding_review_agent_loop/repair.py` should use `_parse_gemini_payload` from `agents.gemini` instead of manually calling `_strip_gemini_preamble` on the raw output. The Gemini CLI can return JSON (e.g., if `--output-format json` is used or in certain environments), and `_parse_gemini_payload` handles this robustly by extracting the inner `response` text. Returning raw JSON from the CLI will cause validation failures in the orchestrator.",
    "The `_strip_gemini_preamble` function in `src/coding_review_agent_loop/agents/gemini.py` should be updated to include `PLAN_STATE_RE` when searching for markers. Currently, it only looks for `STATE_RE` and `CLARIFY_RE`. If a plan review is being repaired (which `_REPAIR_PROMPT` supports), diagnostics appearing before the `AGENT_PLAN_STATE` marker will not be stripped correctly, leading to validation failures."
  ],
  "same_pr_followups": [],
  "future_followups": [
    "Consider restoring token usage logging in `attempt_repair`. Using `_parse_gemini_payload` makes the usage metadata available, allowing for consistency with the previous SDK-based implementation."
  ],
  "prior_item_dispositions": []
}
```


-- Google Gemini

wwind123:
```json
{
  "schema_version": 1,
  "kind": "coder_followup",
  "state": "blocking",
  "summary": "Both blocking items addressed. Switched repair.py to _parse_gemini_payload (handles JSON-wrapped CLI output), and added PLAN_STATE_RE to _strip_gemini_preamble in agents/gemini.py. Two new tests added.",
  "addressed_items": ["item-1"],
  "remaining_items": [],
  "human_requirements": {
    "addressed_ids": [],
    "checked_discussion_directly": false
  },
  "tests_run": ["python3 -m pytest tests/test_agent_loop.py -q  →  429 passed"]
}
```

**Changes in this update (commit `1542702`):**

- `repair.py`: Import `_parse_gemini_payload` instead of `_strip_gemini_preamble`; replace `text = _strip_gemini_preamble(...)` with `text, _, _, _ = _parse_gemini_payload(...)`. This correctly handles JSON-wrapped CLI output that `_strip_gemini_preamble` alone would return as raw JSON.

- `agents/gemini.py`: Added `PLAN_STATE_RE` to the import and to the `marker_matches` list in `_strip_gemini_preamble`, so diagnostics preceding `<!-- AGENT_PLAN_STATE: ... -->` markers are stripped when repairing plan reviews.

- `tests/test_agent_loop.py`: Added `test_attempt_repair_handles_json_wrapped_cli_output` (verifies `_parse_gemini_payload` unwraps JSON CLI output in the repair path) and `test_parse_gemini_output_strips_cli_preamble_before_plan_state_marker` (verifies `PLAN_STATE_RE` preamble stripping).


-- Anthropic Claude

wwind123:
**Review verdict:** Blocking

```json
{
  "schema_version": 1,
  "kind": "pr_review",
  "state": "blocking",
  "summary": "The PR successfully migrates the repair pass to use the Gemini CLI, unifying authentication and removing the `google-genai` dependency. It also addresses the feedback from Round 1 by using `_parse_gemini_payload` and updating the preamble stripping logic. However, the tests are currently fragile because they interact with the host environment's `gemini` CLI, leading to non-deterministic failures in environments where the CLI is installed.",
  "blocking_items": [
    "Tests in `tests/test_agent_loop.py` are not properly isolated from the host environment. When the `gemini` CLI is present in the `PATH`, `attempt_repair` (which is now hardcoded for reviewers) executes real commands during test runs. This causes several existing tests (e.g., `test_usage_summary_keeps_retry_attempts_and_marks_only_validated_call_successful`, `test_pr_loop_retries_plain_agent_state_near_miss_once`) to fail because the real Gemini CLI 'repairs' intentionally malformed test inputs, violating test expectations. `attempt_repair` should be mocked in `tests/test_agent_loop.py` (e.g., via an autouse fixture that returns `None` by default) to ensure deterministic results and environmental isolation."
  ],
  "same_pr_followups": [],
  "future_followups": [],
  "prior_item_dispositions": [
    {
      "item_id": "item-1",
      "disposition": "resolved"
    }
  ]
}
```

- [item-1] Blocking issue from Google Gemini, round 1: ```json -> resolved


-- Google Gemini

wwind123:
```json
{
  "schema_version": 1,
  "kind": "coder_followup",
  "state": "blocking",
  "summary": "Added autouse fixture _no_real_repair that patches coding_review_agent_loop.orchestrator.attempt_repair to return None by default, isolating all orchestrator-path tests from the host Gemini CLI. Consolidated the unittest.mock import to a single top-level statement.",
  "addressed_items": ["item-2"],
  "remaining_items": [],
  "human_requirements": {
    "addressed_ids": [],
    "checked_discussion_directly": false
  },
  "tests_run": ["python3 -m pytest tests/test_agent_loop.py -q  →  429 passed"]
}
```

**Change in this update (commit `c33397d`):**

`tests/test_agent_loop.py`: Added `_no_real_repair` autouse fixture that patches `coding_review_agent_loop.orchestrator.attempt_repair` to return `None` for every test. Explicit repair integration tests (`test_run_pr_loop_uses_repair_pass_on_format_failure`, etc.) override this with their own `patch` context managers, which take precedence. Unit tests for `attempt_repair` itself patch `subprocess.run` directly and are unaffected. Also moved `from unittest.mock import MagicMock, patch` to a single top-level import and removed the duplicate mid-file import.


-- Anthropic Claude

wwind123:
**Review verdict:** Approved

The PR successfully migrates the repair pass to use the Gemini CLI, unifying authentication and removing the `google-genai` dependency. It also addresses the feedback regarding test isolation by mocking `attempt_repair` in a global test fixture. Preamble stripping logic has been enhanced to handle plan state markers.

- [item-2] Blocking issue from Google Gemini, round 2: ```json -> resolved


-- Google Gemini

PR #196, opened 2026-05-31
Fix transient classification for 5xx line refs

### LINKED ISSUE #194: False positive in TRANSIENT_AGENT_OUTPUT_RE: source line numbers matching 5xx HTTP pattern
## Bug

`TRANSIENT_AGENT_OUTPUT_RE` includes `\b5\d\d\b` to detect HTTP 5xx error responses. This pattern also matches three-digit numbers starting with 5 that appear in other contexts — in particular, source file line number references like `orchestrator.py:577-581`.

When an agent response contains such a reference in its content (e.g. in a note field of a structured review), `_is_transient_agent_output` returns `True`, which:
- Causes the orchestrator to treat the failure as a transient retry rather than a validation error
- Bypasses the repair pass entirely (`use_repair and not _is_transient_agent_output(text)` is False)
- Results in 3 identical retries, all failing the same way, before surfacing the error

## Observed instance

Plan review for issue #186, round 2. Claude produced a valid `plan_review` JSON payload wrapped in a code fence (the underlying format issue). The response included this text in a `prior_plan_item_dispositions` note:

```
"orchestrator.py lines 577-581: it currently falls back to parse_plan_state(text)"
```

`577` matched `\b5\d\d\b`. The orchestrator retried 3 times instead of triggering the repair pass, then exited with "Failure category: transient (rerun may succeed)" — which is incorrect.

The actual failure was deterministic: Claude consistently produces JSON-in-fence format, which the markdown fallback silently accepts while dropping the disposition data. That is addressed separately by #186.

## Fix

Narrow `\b5\d\d\b` so it only matches HTTP status codes, not arbitrary three-digit numbers. Options:

**Option A** — require HTTP context:
```python
r"HTTP/[\d.]+ 5\d\d|status[_ ]?code[: ]+5\d\d|\b5\d\d\b(?= [A-Z])"
```

**Option B** — exclude file:line patterns with a negative lookbehind:
```python
r"(?<!:\d)(?<!\w)5\d\d(?!\d)"  # still fragile
```

**Option C (simplest)** — replace `\b5\d\d\b` with explicit status codes or a more specific prefix:
```python
r"(?:HTTP[/ ]\S+ )?5(?:00|02|03|04|22|02|03)\b"
```

**Option D** — drop the generic `5\d\d` pattern entirely and rely on the already-present explicit patterns (`Internal Server Error`, `Bad Gateway`, `Service Unavailable`, `Gateway Timeout`, `overloaded`) which cover the real transient cases without false positives. The `5\d\d` catch-all adds marginal coverage while creating this class of false match.

Option D is the lowest risk and is recommended: the explicit text patterns already cover the HTTP 5xx cases that matter in practice.

## Impact

When this false positive triggers:
- The repair pass is skipped for what is actually a deterministic format error
- 3 retries waste time and tokens producing identical failures
- The error is misclassified as "transient (rerun may succeed)" when the actual cause is a format issue requiring the repair pass or #186

-- Anthropic Claude

wwind123:
## Summary
- Remove the generic bare 5xx number match from transient agent output classification.
- Add regressions proving source line references like orchestrator.py:577-581 are deterministic, while explicit server-error phrases remain transient.
- Add repair-path coverage showing malformed review output with a 5xx-looking source line still invokes repair.

Fixes #194

## Tests
- python -m pytest tests/test_agent_loop.py -k "transient or repair" (failed: python command not found)
- python3 -m pytest tests/test_agent_loop.py -k "transient or repair" (passed)
- python3 -m pytest (passed)

### DIFF
--- a/src/coding_review_agent_loop/orchestrator.py
+++ b/src/coding_review_agent_loop/orchestrator.py
@@ -177,7 +177,7 @@ TRANSIENT_AGENT_OUTPUT_RE = re.compile(
     r"Invalid stream|empty response|malformed tool call|"
     r"network (?:reset|timeout)|connection (?:reset|timed out|timeout)|"
     r"\btimed out\b|\btimeout\b|"
-    r"\b5\d\d\b|Internal Server Error|Bad Gateway|Service Unavailable|Gateway Timeout|"
+    r"Internal Server Error|Bad Gateway|Service Unavailable|Gateway Timeout|"
     r"\b429\b|rate.?limit(?:ed)?|"
     r"session.?limit.?exceeded|session_limit_exceeded|too many sessions|"
     r"no capacity available|capacity.*(?:unavailable|exceeded)|"
--- a/tests/test_agent_loop.py
+++ b/tests/test_agent_loop.py
@@ -40,6 +40,8 @@ from coding_review_agent_loop.cli import (
 from coding_review_agent_loop.errors import QuotaResetExceededError
 from coding_review_agent_loop.orchestrator import (
     _format_reset_duration,
+    _failure_category,
+    _is_transient_agent_output,
     _parse_rate_limit_reset_seconds,
 )
 from coding_review_agent_loop.config import (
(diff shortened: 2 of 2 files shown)


wwind123:
Implemented the approved fix for issue #194 and opened PR #196:

- Removed the generic bare 5xx transient-output match so source references like `orchestrator.py:577-581` classify as deterministic.
- Added regression coverage for 5xx-looking source line references, explicit server-error phrases, and the repair path.

Tests: `python -m pytest tests/test_agent_loop.py -k "transient or repair"` (failed: `python` not found); `python3 -m pytest tests/test_agent_loop.py -k "transient or repair"` (passed); `python3 -m pytest` (passed).





-- OpenAI Codex

wwind123:
**Review verdict:** Approved

Fix correctly removes the overly-broad 5xx pattern that was falsely matching file:line references, while preserving explicit HTTP error detection. Comprehensive test coverage validates the fix.


-- Anthropic Claude

wwind123:
**Review verdict:** Approved

The PR successfully fixes the false positive transient classification for 5xx line references by removing the overly broad `\b5\d\d\b` pattern from `TRANSIENT_AGENT_OUTPUT_RE`. It correctly implements Option D from issue #194, relying on explicit HTTP error phrases which are already present in the regex. The added tests verify that line numbers no longer trigger retries while actual server error messages still do.


-- Google Gemini

PR #84, opened 2026-05-19
Sync coder checkout before implementation

wwind123:
## Summary
- add a just-in-time coder base-branch sync helper for issue/task implementation flows
- sync before issue implementation, task implementation, and plan-first implement-after-approval
- preserve PR mode branch state and dirty explicit workdir failures

## Tests
- python3 -m pytest

PR #69, opened 2026-05-17
Include issue comments in issue-mode context

wwind123:
## Summary
- fetch issue title, body, and comments for issue-mode runs
- include bounded issue context in coder prompts and issue-origin review prompts
- document issue comment handling and add regression coverage

Fixes #68

## Tests
- /tmp/coding-review-agent-loop/scratch/issue68-venv/bin/python -m pytest -q

PR #218, opened 2026-06-02
Reconcile approved follow-up publishing

### LINKED ISSUE #85: Reconcile approved follow-ups across PR rounds before filing issues
## Feature request

Improve approved-followup reconciliation so useful follow-ups are not lost, while avoiding stale/noisy issue creation.

## Background

The tool has already hit both sides of this trade-off:

1. PR #69 created stale follow-up issues from earlier approved reviews even after those items were fixed in later rounds. Issue #73 / PR #75 addressed that by creating follow-up issues only from the final approval round.
2. PR #84 showed the opposite failure mode: Gemini listed useful future follow-ups in round 1, but did not repeat them in the final approval round. Because only final-round follow-ups are now processed, no follow-up issues were filed.

So the current final-round-only heuristic reduces noise but can lose real follow-ups.

## Problem

Regex extraction from final-round comments is too brittle to capture the desired intent:

- Earlier future follow-ups may remain valid but disappear if a reviewer forgets to restate them.
- Earlier follow-ups may become stale after the coder fixes them.
- Multiple reviewers may describe the same follow-up differently.
- Some same-PR-worthy items may be miscategorized as future work, or vice versa.

## Possible approach

Consider adding an explicit reconciliation step before creating approved-followup issues.

Options:

1. Rule-based reconciliation:
   - Collect all future follow-ups from all approved rounds.
   - Ask the coder to mark which ones were addressed in later commits.
   - Only file unresolved items.

2. Reviewer restatement:
   - In final review prompts, include previously collected future follow-ups and ask reviewers to explicitly restate which ones remain valid.

3. LLM-based reconciliation:
   - After final approval, call an agent to review:
     - all previously collected future follow-ups,
     - coder update comments,
     - final reviewer comments,
     - PR diff/commits if needed,
   - Output a deduplicated list of follow-ups that still need separate tracking.

The LLM-based approach may be more robust than regex-only logic, but should be carefully constrained and signed. The output should probably be posted as a `-- coding-review-agent-loop` summary if produced by orchestration rather than by a named reviewer.

## Design constraints

- Avoid filing stale issues for already-fixed items.
- Avoid losing valid earlier follow-ups that were not repeated in the final round.
- Deduplicate similar follow-ups across reviewers.
- Keep issue creation bounded, e.g. at most three follow-up issues unless explicitly configured otherwise.
- Make it clear in logs/comments why a follow-up was filed, skipped, or deduplicated.

## Concrete examples

- PR #69: stale #70/#71 were filed even though they had been fixed before final approval.
- PR #84: useful Gemini round-1 follow-ups about memory freshness and sync-before-planning were not filed because they were not repeated in the final round.

-- OpenAI Codex

wwind123:
## Summary
- reconcile approved-review future follow-ups from the remaining unresolved-item ledger before publishing
- group semantic duplicates deterministically while preserving reviewer provenance and disposition notes
- update prompts, docs, and regression coverage for stale, missing, duplicate, and capped follow-up cases

Fixes #85

## Tests
- python3 -m compileall -q src tests
- git diff --check
- python3 -m pytest

### DIFF
--- a/README.md
+++ b/README.md
@@ -323,8 +323,17 @@ larger redesigns and independent work belong under Future follow-ups. Approved
 future follow-ups remain in the round-to-round
 ledger so later reviewers can explicitly confirm they are still future work,
 resolved, or should be promoted back to same-PR or blocking status. The final
-summary or issue creation uses the remaining future items from that ledger. The
-issue modes create at most three follow-up issues to avoid issue noise.
+summary or issue creation uses the remaining future items from that reconciled
+ledger, not only the final round's newly written Future follow-ups.
+
+Before posting summaries or creating issues, the loop deduplicates the remaining
+future items across reviewers using deterministic topic keys from headings,
+code identifiers, docs/files, and normalized wording. The selected issue body
+keeps the canonical wording plus an `Original reviewer notes` section so
+reviewer provenance and later disposition notes are not lost. The issue modes
+create at most three follow-up issues to avoid issue noise, and the final PR
+comment reports how many items were filed or summarized, deduplicated, or
+skipped by the cap.
 
 Plan reviews follow the same rule: approved plan reviews may include Future
 follow-ups only. Blocking plan issues, Same-plan follow-ups, or carried-forward
--- a/src/coding_review_agent_loop/followups.py
+++ b/src/coding_review_agent_loop/followups.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import re
+from collections import Counter
 from collections.abc import Sequence
 from dataclasses import dataclass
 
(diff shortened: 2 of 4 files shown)


wwind123:
Implemented the approved plan and opened PR #218:

https://github.com/wwind123/coding-review-agent-loop/pull/218

Tests: `python3 -m compileall -q src tests` passed; `git diff --check` passed; `python3 -m pytest` passed.




-- OpenAI Codex

wwind123:
**Review verdict:** Approved

Implements deterministic, ledger-aware reconciliation for approved future follow-ups: key-based + Jaccard-similarity deduplication over all unresolved ledger items solves both failure modes from issue #85 (stale items filed from PR #69 pattern; useful items lost as in PR #84 pattern). All 516 tests pass, including 5 new targeted regression tests.


-- Anthropic Claude

wwind123:
**Review verdict:** Approved

This PR improves the reconciliation of approved follow-ups across review rounds. It introduces semantic deduplication to group similar follow-ups, ensures earlier follow-ups are preserved unless explicitly resolved, and provides better reporting on the reconciliation process. These changes directly address the issues of lost follow-ups and stale issue creation identified in #85.


-- Google Gemini

PR #293, opened 2026-06-12
fix: detect .git-only stale workdir; wire run_external repo cloning

wwind123:
## Summary

- **`_is_stale_default_workdir` extended** (`config.py`): previously returned `False` as soon as `.git/` existed, so a workdir with `.git/` but no source files (e.g. after accidental wipe leaving only response storage at `.git/agent-loop/`) was never detected as stale. Now also flags the case where `.git` is present but the working tree contains no non-artifact files.

- **`run_external.py` now validates/re-clones the workdir** (`run_external.py`, `skill_runner.py`): `run_external` was being called with `--workdir` but never called `ensure_temp_checkout`, so it would fail silently if the workdir was absent or stale. Added `--repo OWNER/REPO`; when provided, `run_external` calls `ensure_temp_checkout` with the real repo before launching the agent. `skill_runner` now passes `--repo` on every `run_external` invocation.

- Fixed a misleading comment in `skill_runner.py` that claimed cloning was handled by `--workdir` (it wasn't).

## Test plan

- [ ] `tests/test_agent_loop.py -k stale` — three tests, including new `test_stale_default_workdir_git_only_is_recreated`
- [ ] `tests/test_skill_helpers.py` — 22 tests unchanged

-- Anthropic Claude

### DIFF
--- a/helpers/run_external.py
+++ b/helpers/run_external.py
@@ -147,9 +147,9 @@ def main() -> None:
     # Import backends lazily to avoid heavy import in dry-run path
     from coding_review_agent_loop.agents.codex import CodexBackend
     from coding_review_agent_loop.agents.gemini import GeminiBackend
-    from coding_review_agent_loop.config import AgentLoopConfig, ensure_temp_checkout
+    from coding_review_agent_loop.config import AgentLoopConfig, AgentName, ensure_temp_checkout
 
-    agent_name = args.agent
+    agent_name: AgentName = args.agent
     default_cmds = {"codex": "codex", "gemini": "gemini"}
     cmd = args.cmd or default_cmds[agent_name]
 
@@ -164,7 +164,7 @@ def main() -> None:
         codex_dir=workdir,
         gemini_dir=workdir,
         coder="claude",
-        reviewer=(agent_name,),  # type: ignore[arg-type]
+        reviewer=(agent_name,),
         base="main",
         max_rounds=1,
         auto_merge=False,
@@ -191,12 +191,12 @@ def main() -> None:
         refresh_agent_memory=False,
         agent_memory_dir=log_dir,
         refresh_test_profile=False,
-        auto_agent_dirs=(agent_name,),  # type: ignore[arg-type]
+        auto_agent_dirs=(agent_name,),
     )
 
     runner = Runner(dry_run=False)
     if args.repo:
-        ensure_temp_checkout(workdir, agent=agent_name, config=config, runner=runner)  # type: ignore[arg-type]
+        ensure_temp_checkout(workdir, agent=agent_name, config=config, runner=runner)
     backend = CodexBackend() if agent_name == "codex" else GeminiBackend()
     try:
         result = backend.run(runner, config, prompt)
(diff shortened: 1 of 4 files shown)


wwind123:
**Review verdict:** Blocking

The PR correctly extends stale workdir detection to handle .git-only directories and wires up repository cloning in `run_external.py`. However, it fails to handle completely empty directories as stale, which causes `skill_runner.py` to fail on first run because it pre-creates the workdir.

### Blocking issues
- `_is_stale_default_workdir` in `src/coding_review_agent_loop/config.py` still requires the directory to be non-empty (via `bool(contents)`) to be considered stale in the non-git case. This blocks auto-cloning when `helpers/skill_runner.py` creates an empty directory before calling `run_external.py`, as `ensure_temp_checkout` will find the directory exists but is not stale, skipping the clone and failing later with a 'not a git checkout' error.

### Same-PR follow-ups
- Import `AgentName` in `helpers/run_external.py` and remove the `type: ignore[arg-type]` on the `ensure_temp_checkout` call to maintain type safety.


-- Google Gemini

wwind123:
### Change summary (round 1 feedback)

**Gemini blocking item (real bug, fixed):** `_is_stale_default_workdir` required `bool(contents)` to be true in the no-`.git` branch, so an empty directory was never flagged as stale. When `ensure_temp_checkout` found an empty-but-existing Codex workdir, it skipped the clone and crashed in `_sync_base_branch`. Fixed by removing the `bool(contents)` guard — an empty directory with no git checkout is always stale.

**Gemini same-PR followup (fixed):** `run_external.py` now declares `agent_name: AgentName` (cast from the argparse string, which is always valid) and removes all three `type: ignore[arg-type]` annotations.

Added `test_stale_default_workdir_empty_is_recreated` covering the empty-dir case. All 4 stale workdir tests pass.

-- Anthropic Claude

wwind123:
**Review verdict:** Approved

The PR successfully addresses the prior review items and improves the robustness of workdir management in `run_external`. `_is_stale_default_workdir` now correctly identifies empty or .git-only directories as stale, triggering a re-clone when necessary. The addition of `--repo` to `run_external` ensures that it can autonomously manage the checkout state. Type safety is also improved by properly importing and using `AgentName`.

- [item-1] Blocking issue from Google Gemini, round 1: `_is_stale_default_workdir` in `src/coding_review_agent_loop/config.py` still requires the direct... -> resolved
- [item-2] Same-PR follow-up from Google Gemini, round 1: Import `AgentName` in `helpers/run_external.py` and remove the `type: ignore[arg-type]` on the `e... -> resolved


-- Google Gemini

wwind123:
**Review verdict:** Approved

Approved. The stale workdir detection now covers empty and .git-only default workdirs, and run_external is wired to validate or re-clone when skill_runner supplies the repo.

- [item-1] Blocking issue from Google Gemini, round 1: `_is_stale_default_workdir` in `src/coding_review_agent_loop/config.py` still requires the direct... -> resolved
- [item-2] Same-PR follow-up from Google Gemini, round 1: Import `AgentName` in `helpers/run_external.py` and remove the `type: ignore[arg-type]` on the `e... -> resolved


-- OpenAI Codex

Issue #296, opened 2026-06-13
Increment 2: test-command gate in run-pr-round (#294)


wwind123:
## Summary

Increment 2 of #294 (close the CLI↔skill automation gap). Add a **test gate** to
the skill so a PR review round can run the project's test command and report
pass/fail, instead of leaving "did the tests actually pass?" to the host's memory.

The CLI has `--test-command` / `--pre-review-tests` (run via
`coding_review_agent_loop.checks`); the skill has nothing equivalent. This is a
mechanical, deterministic gate — exactly the kind of work that should be code,
not host bookkeeping.

## Scope

- Applies to **`run-pr-round` only**. A plan round reviews a document; there is
  no code to test until a PR exists.
- The gate **reports**, it does not silently merge. Merge stays a human decision
  (per #294).

## Proposed behavior

- New flags on `run-pr-round`:
  - `--test-command CMD` (optional; when omitted, no gate, current behavior).
  - `--test-workdir DIR` (default `.`) — where to run it. The host coder has the
    PR branch checked out locally; the reviewers' temp clones are not the target.
- Run the command after the reviewer turns, capturing exit code (do **not** raise
  / crash the round). Add to the JSON result:
  ```json
  "tests": { "command": "...", "passed": true, "exit_code": 0 }
  ```
- `state` stays reviewer-driven; "ready to merge" = `state == "approved"` **and**
  `tests.passed` (documented in `SKILL.md`).

## Open questions for reviewers

- Should a test failure force `state: blocking`, or only populate `tests` and let
  the documented host gate handle it? (Leaning toward the latter — keep `state`
  meaning "reviewer consensus", report tests separately.)
- Capture and surface a tail of test output in the result, or just the exit code?

-- Anthropic Claude

wwind123:
# Plan: test-command gate in run-pr-round (issue #296, increment 2 of #294)

## Goal & scope

Add an optional, deterministic test gate to the skill's PR review round so a round
can run the project's test command and **report** pass/fail in its JSON result.
Merge stays a human decision (#294); the gate never merges and never crashes the
round.

- **`run-pr-round` only.** A plan round reviews a document — no code exists to
  test until a PR is open. `run-plan-round` is unchanged.
- Opt-in: with no `--test-command`, behavior is identical to today.

## Decisions (resolving the issue's open questions)

1. **A test failure does NOT flip `state`.** `state` keeps its current meaning —
[...]

Issue #294, opened 2026-06-13
Close CLI↔skill automation gap (merge stays human)


wwind123:
## Summary

The Claude Code skill (`helpers/skill_runner.py` + `SKILL.md`) lags the CLI
(`orchestrator.py`) on automation. The skill is a **single-round primitive**
(`run-plan-round`, `run-pr-round`, `retry-validate`) with the host driving the
loop by hand via `SKILL.md`; the CLI is a **closed-loop driver** that loops to
approval, implements, tests, waits for CI, and merges — unattended.

Goal: **close the automation gap as much as possible**, with one deliberate
exception — **PR merge stays a human decision** (the skill is positioned as
interactive-session usage; unattended merge is the line we don't cross).

## Framing: determinism vs. judgment (not skill vs. host)

"The host owns it" is not a reason to leave something un-automated — when the
host follows `SKILL.md` that *is* automation, just LLM-driven. The right axis is:

- **Deterministic / mechanical → code (skill_runner).** An LLM doing bookkeeping
  is strictly worse than code: it can miscount rounds, skip a gate, mis-order
  save-before-emit, and it burns context every round. This is *why* skill_runner
  already owns the round pipeline.
- **Judgment / context-dependent → host.** Writing/revising the plan, the actual
  coding, deciding to clarify. Already automated as host loops; can't be coded.

By that test, several "host owns it" items are actually **under-automated**:
they're neither coded nor reliably scripted in `SKILL.md`, so they just don't
happen.

## Parity gaps (CLI has, skill lacks)

Mechanical (should become code):
- **Transient-failure retries** — `run_external` hardcodes `agent_max_retries=0`;
  CLI default is 2 with backoff + rate-limit/quota classification.
- **Test gates** — CLI `--test-command` / `--pre-review-tests`; skill has none.
- **Approved-followups publishing** — CLI has 5 modes
  (`ignore`/`summarize`/`issue`/`fix-and-summarize`/`fix-and-issue`); the skill
  surfaces followups in JSON but never acts on them.
- **Loop scaffolding** — CLI loops to approval / `--max-rounds`; skill relies on
  the host re-reading `SKILL.md` each round. (Note: a Python `run-loop` can't own
  the *whole* loop because the coder **is** the host session — so the realistic
  win is folding the mechanical gates into the round commands + tightening the
  `SKILL.md` loop, not a subprocess loop.)
- **Response recovery** — CLI auto-recovers near-miss structured JSON; skill only
[...]

PR #298, opened 2026-06-13
Fix reviewer prompts embedding wrong checkout path (#297)

### LINKED ISSUE #297: Reviewer prompts embed wrong checkout path (skill-runner vs skill-runner-{agent}) → false Codex blocks
## Summary

The skill embeds the **wrong checkout path** in reviewer prompts, so reviewers
are told to inspect a directory that does not exist. Reviewers that insist on
local inspection (Codex, consistently) hard-block with "No such file or
directory"; reviewers that review from the plan/diff text (Gemini) don't notice.
This is the root cause of the recurring "Codex checkout is messed up" /
"assigned checkout was not available" blocks.

## Root cause

- Actual reviewer workdir — `helpers/skill_runner.py` `_workdir_for_agent` (and
  `run_external`): `/tmp/coding-review-agent-loop/skill-runner-{agent}`
  (e.g. `skill-runner-codex`).
- Path used to **build the prompt** — `helpers/prompt_builders.py`
  `_make_minimal_config`: hardcodes `claude_dir`/`codex_dir`/`gemini_dir` all to
  `/tmp/coding-review-agent-loop/skill-runner` (no `-{agent}` suffix).

The prompt-building config's dirs never match the real per-agent workdirs, so the
"your checkout is at X" text in every reviewer prompt points at a nonexistent
path. It also collapses all three agents onto one shared dir, which is wrong in
its own right.

## Evidence

On issue #296 round 2, Codex blocked solely with:
> Required local-context inspection could not be completed because the assigned
> checkout path `/tmp/coding-review-agent-loop/skill-runner` was not available.

while dispositioning all six prior plan items as `resolved`. The actual checkout
`/tmp/coding-review-agent-loop/skill-runner-codex` was present and current.

## Proposed fix

Thread each reviewer's **actual** workdir into prompt construction so the
embedded path matches where the agent really runs:
- `prompt_builders.build_plan_review_prompt_for_skill` / the PR equivalent should
  accept the reviewer's workdir (the value `_workdir_for_agent` returns, honoring
  `--workdir` / `--workdir-{agent}` overrides) and set that agent's dir on the
  config used for prompt building — rather than `_make_minimal_config` hardcoding
  `skill-runner`.
- Keep the other agents' dirs irrelevant to the prompt (only the active
  reviewer's path is embedded).

## Tests

- Unit-test that the built prompt for a given reviewer + workdir embeds that exact
  workdir path (and the `-{agent}` suffix), not the bare `skill-runner`.

## Impact

High signal-to-noise: fixes false Codex blocks on every review round and removes
a class of wasted review cycles. Small, contained change.

-- Anthropic Claude

wwind123:
## What & why

Fixes #297. The skill embedded the **wrong checkout path** in every reviewer
prompt: `prompt_builders._make_minimal_config` hardcoded
`/tmp/coding-review-agent-loop/skill-runner` (no agent suffix), but reviewers
actually run in `/tmp/coding-review-agent-loop/skill-runner-{agent}` (per
`skill_runner._workdir_for_agent`).

So reviewers were told to inspect a directory that doesn't exist. Reviewers that
insist on local inspection (Codex, consistently) blocked with "No such file or
directory"; reviewers that work from the plan/diff text (Gemini) didn't notice.
This is the root cause of the recurring "reviewer checkout is messed up" blocks
— including the false Codex block on #296 round 2, where Codex resolved all six
plan items but blocked solely on the bogus path.

## Changes

- `prompt_builders._make_minimal_config` derives each agent's dir from the real
  `skill-runner-{agent}` pattern and uses the active reviewer's **actual**
  workdir when provided, instead of the bare `skill-runner` path.
- `build_plan_review_prompt_for_skill` / `build_review_prompt_for_skill` accept a
  `workdir` argument; both `skill_runner` call sites pass the value from
  `_workdir_for_agent` (honoring `--workdir` / `--workdir-{agent}` overrides),
  computed before prompt construction.

## Tests

- New `TestPromptCheckoutPath` asserts the built prompt embeds the reviewer's
  real workdir and never the bare `skill-runner` path — for explicit-workdir,
  default-workdir, and PR flows.
- Full suite: **752 passed**.

## Note

This is a prerequisite uncovered while reviewing increment 2 (#296) of the
automation-gap work (#294); it unblocks clean Codex reviews for everything after.

### DIFF
--- a/helpers/prompt_builders.py
+++ b/helpers/prompt_builders.py
@@ -25,13 +25,28 @@ def _make_minimal_config(
     repo: str,
     coder: AgentName,
     reviewer_names: Sequence[AgentName],
+    *,
+    reviewer: AgentName | None = None,
+    workdir: str | None = None,
 ) -> AgentLoopConfig:
-    tmp = Path(tempfile.gettempdir()) / "coding-review-agent-loop" / "skill-runner"
+    base = Path(tempfile.gettempdir()) / "coding-review-agent-loop"
+    legacy = base / "skill-runner"
+
+    def _agent_dir(agent: AgentName) -> Path:
+        # The active reviewer's dir must match where the agent actually runs, so the
+        # checkout path embedded in the prompt is real. The default mirrors
+        # skill_runner._workdir_for_agent ("skill-runner-{agent}"); previously this
+        # hardcoded the bare "skill-runner" path, which never exists and made
+        # reviewers (notably Codex) block on a missing checkout (#297).
+        if workdir and agent == reviewer:
+            return Path(workdir)
+        return base / f"skill-runner-{agent}"
+
     return AgentLoopConfig(
         repo=repo,
-        claude_dir=tmp,
-        codex_dir=tmp,
-        gemini_dir=tmp,
+        claude_dir=_agent_dir("claude"),
+        codex_dir=_agent_dir("codex"),
+        gemini_dir=_agent_dir("gemini"),
         coder=coder,
         reviewer=tuple(reviewer_names),
         base="main",
@@ -52,13 +67,13 @@ def _make_minimal_config(
         ci_timeout_seconds=300,
         ci_poll_interval_seconds=30,
         quiet=False,
-        log_dir=tmp,
+        log_dir=legacy,
         progress_interval_seconds=30,
         agent_max_retries=0,
         agent_retry_backoff_seconds=(30,),
         agent_memory=False,
         refresh_agent_memory=False,
-        agent_memory_dir=tmp,
+        agent_memory_dir=legacy,
         refresh_test_profile=False,
         auto_agent_dirs=tuple(reviewer_names),
     )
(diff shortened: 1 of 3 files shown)


wwind123:
**Review verdict:** Approved

The PR correctly addresses the issue where reviewer prompts embedded an incorrect hardcoded checkout path. By deriving the agent-specific directories and allowing the actual workdir to be passed during prompt construction, it ensures that reviewers receive accurate instructions for local inspection. The added tests effectively verify that the bare `skill-runner` path is no longer used in prompts.


-- Google Gemini

wwind123:
**Review verdict:** Approved

The prompt builders now embed the active reviewer's actual checkout path, and the skill runner passes the same workdir used for invocation. The added tests cover explicit and default reviewer checkout paths for plan and PR prompts.


-- OpenAI Codex

PR #342, opened 2026-06-17
Fix workdir guard for script tag regexes

### LINKED ISSUE #341: Workdir guard misreads JS regex closing tags as absolute paths
The coder test-path guard can falsely reject valid in-checkout test reports when a reported command contains a JavaScript/HTML regex with a closing tag such as `</script>`.

Observed while running:

```sh
~/tools/coding-review-agent-loop/.venv/bin/agent-loop pr 243 --repo wwind123/llm-dialectic --coder codex --reviewer claude --reviewer antigravity --dangerous-agent-permissions --approved-followups fix-and-issue --max-rounds=10
```

Failure:

```text
agent-loop: Coder reported tests from outside the assigned checkout: '/script' in command 'node -e "const fs=require(\'fs\'); const html=fs.readFileSync(\'server/static/index.html\',\'utf8\'); const scripts=[...html.matchAll(/<script(?![^>]*\\\\bsrc=)[^>]*>([\\\\s\\\\S]*?)<\\\\/script>/gi)].map(m=>m[1]); scripts.forEach((code,i)=>{ try { new Function(code); } catch(e) { console.error(\'script \'+i+\' parse failed\'); throw e; } }); console.log(scripts.length+\' inline scripts parsed\');" (failed: naive regex matched non-code text)'. Assigned checkout: /tmp/coding-review-agent-loop/wwind123-llm-dialectic/codex/repo
```

The reported command did not execute from `/script`. The guard's broad absolute-path regex appears to match the `/script` substring inside the JavaScript regex literal `<\/script>` / closing-tag pattern. That turns a quoted command snippet into a bogus filesystem path and aborts the loop.

Evidence from the same Codex run:

- Codex worked in the assigned checkout: `/tmp/coding-review-agent-loop/wwind123-llm-dialectic/codex/repo`.
- It committed and pushed `23093a0 fix: handle hidden-tab debate rejection`.
- It wrote a structured public response.
- The rejection was only caused by the test report containing the failed naive `node -e` syntax-check command.

Expected behavior:

- The workdir guard should continue rejecting explicit absolute/home/Windows paths that identify real command paths or working directories outside the assigned checkout.
- It should not treat HTML closing tags or JavaScript regex fragments inside quoted snippets, such as `</script>` or `<\/script>`, as filesystem paths.

Potential approaches:

- Tokenize commands first and inspect command/path-like tokens instead of applying `ABSOLUTE_PATH_RE` across the entire raw command string.
- Or suppress absolute-path matches that are clearly part of HTML/XML closing tags, e.g. preceded by `<`, or escaped regex closing tags in quoted code snippets.
- Add a regression test using the exact `node -e` command above.

This should be a narrow fix; the existing outside-checkout protections should stay intact.

-- Human Reviewer

wwind123:
Summary:
- Ignore absolute-path regex matches that are part of HTML/JS closing-tag regex syntax.
- Add a regression test for the reported inline script parser command.

Tests:
- python3 -m pytest tests/test_agent_loop.py -k workdir_guard
- python3 -m pytest

Fixes #341

### DIFF
--- a/src/coding_review_agent_loop/workdir_guard.py
+++ b/src/coding_review_agent_loop/workdir_guard.py
@@ -11,7 +11,7 @@ from .errors import AgentLoopError
 
 
 TEST_SECTION_RE = re.compile(r"(?im)^\s*tests(?:\s+run)?\s*:\s*(?P<body>.*)$")
-ABSOLUTE_PATH_RE = re.compile(r"(?<![\w.-])/(?:[^\s`'\"|;&)<>]+)")
+ABSOLUTE_PATH_RE = re.compile(r"(?<![\w.\\<>-])/(?:[^\s`'\"|;&)<>]+)")
 HOME_PATH_RE = re.compile(r"(?<![\w.-])(?:~|\$HOME)/(?:[^\s`'\"|;&)<>]+)")
 WINDOWS_PATH_RE = re.compile(r"(?<![\w.-])[A-Za-z]:\\[^\s`'\"|;&)<>]+")
 
--- a/tests/test_agent_loop.py
+++ b/tests/test_agent_loop.py
@@ -919,6 +919,18 @@ def test_workdir_guard_accepts_assigned_absolute_path(tmp_path):
     )
 
 
+def test_workdir_guard_accepts_javascript_regex_closing_script_tag(tmp_path):
+    assigned = tmp_path / "codex" / "repo"
+    assigned.mkdir(parents=True)
+
+    validate_test_commands_within_workdir(
+        (
+            r"""node -e "const fs=require('fs'); const html=fs.readFileSync('server/static/index.html','utf8'); const scripts=[...html.matchAll(/<script(?![^>]*\\bsrc=)[^>]*>([\\s\\S]*?)<\\/script>/gi)].map(m=>m[1]); scripts.forEach((code,i)=>{ try { new Function(code); } catch(e) { console.error('script '+i+' parse failed'); throw e; } }); console.log(scripts.length+' inline scripts parsed');" (failed: naive regex matched non-code text)""",
+        ),
+        assigned_workdir=assigned,
+    )
+
+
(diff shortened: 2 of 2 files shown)


wwind123:
Opened PR #342 against `main`: https://github.com/wwind123/coding-review-agent-loop/pull/342

Implemented a narrow workdir guard fix so broad absolute-path scanning no longer treats slashes in HTML/JavaScript closing-tag regex syntax, such as `<\/script>` and the following `/gi` flags delimiter, as filesystem paths. Added regression coverage using the reported `node -e` command.

Tests: `python -m pytest tests/test_agent_loop.py -k workdir_guard` failed because `python` is not installed; `python3 -m pytest tests/test_agent_loop.py -k workdir_guard` passed; `python3 -m pytest` passed.



### Human requirements
- Requirement 1: Addressed by narrowing the absolute-path regex so script-tag regex fragments are ignored while existing token-based outside-checkout checks remain, and by adding a regression test with the reported `node -e` command.





-- OpenAI Codex

wwind123:
**Review verdict:** Approved

Narrow, correct fix: adding `<`, `>`, and `\` to the ABSOLUTE_PATH_RE negative lookbehind suppresses false matches on HTML/XML closing tags and escaped slash sequences in JS regex literals. All three signed human requirements are met: the false rejection is fixed, real absolute paths outside the checkout are still caught, and a regression test with the exact failing command is included. All 749 tests pass locally and GitHub CI is green.



-- Anthropic Claude: claude-haiku-4-5-20251001

wwind123:
**Review verdict:** Approved

The PR correctly addresses the issue where the workdir guard falsely identified JavaScript/HTML regex fragments (like `</script>`) as absolute paths. It adds a narrow lookbehind to the absolute path regex to exclude these cases and includes a regression test verifying the fix. All tests passed, and the fix is focused and effective.



-- Google Gemini

PR #386, opened 2026-06-19
fix: reword Antigravity output marker instruction to require response before turn ends

### LINKED ISSUE #385: fix: Antigravity ends turn without writing response when it launches background test commands
## Problem

Antigravity (Gemini 3.5 Flash via `agy`) consistently produces empty/narration-only responses on PR reviews when it launches test commands. The current instruction in \`antigravity.py\` says:

> "Do not print the marker until you are done with all internal reasoning, tool use, and review work."

Antigravity interprets this as "I must wait for my background test command to finish before printing the marker." When it schedules a test run (e.g. \`pytest\`) and then ends its turn with "I will wait for the test command to execute and complete", the turn ends before the marker is printed and the response file is empty.

The repair pass also fails because there is no JSON content in the output to extract — just narration.

## Root cause

The wording in \`_with_public_response_marker_instruction\` teaches the wrong behavior: "don't print until fully done" → agent waits → turn ends → empty response.

## Fix

In \`src/coding_review_agent_loop/agents/antigravity.py\`, change the marker instruction to explicitly require the marker/response to be written BEFORE the turn ends, even if background verification steps were started but not completed.

In \`src/coding_review_agent_loop/agents/base.py\`, add a note to \`with_public_response_file_instruction\` that the response file must be written before the turn ends.

## Observed behavior

```
I will run the sitemap tests to verify they pass locally.
I will wait for the test command to execute and complete.
```
(turn ends — response file empty — validation fails)

## Expected behavior

Antigravity writes its review JSON based on code analysis it has already completed, then optionally runs tests — or at minimum writes the response before ending its turn regardless of pending background work.

-- Anthropic Claude

wwind123:
## Summary

- Replaces "Do not print the marker until you are done with all internal reasoning, tool use, and review work" with explicit "print marker → output response → end turn immediately — no further tool calls, narration, or output" sequencing
- Adds "do not defer to a background task result" to handle the specific case where Antigravity launches a background test command and ends the turn waiting for it
- Adds a must-write-before-turn-ends note to the shared response-file instruction in `base.py`

## Root cause

Antigravity (Gemini 3.5 Flash via `agy`) consistently ended turns empty when reviewing complex PRs: it would run `pytest` in the background, then end with "I will wait for the test command to execute and complete", never printing the public-response marker or writing the response file. The old instruction's "don't print until done" wording actively encouraged this.

## Test plan

- [ ] `test_antigravity_prompt_includes_terminal_response_instruction`: asserts new "end your turn immediately" and "do not defer to a background task result" wording is present
- [ ] `test_antigravity_prompt_excludes_old_wait_instruction`: asserts old "Do not print the marker until you are done with all internal reasoning" is absent
- [ ] `test_base_response_file_instruction_includes_must_write_before_turn_ends`: asserts base instruction includes "before your turn ends"
- [ ] Full suite: 947 passed

Closes #385

### DIFF
--- a/src/coding_review_agent_loop/agents/antigravity.py
+++ b/src/coding_review_agent_loop/agents/antigravity.py
@@ -55,8 +55,14 @@ GitHub, print this exact line immediately before it:
 
 {PUBLIC_RESPONSE_MARKER}
 
-Only content after that line will be posted to GitHub. Do not print the marker
-until you are done with all internal reasoning, tool use, and review work.
+Only content after that line will be posted to GitHub. Run any verification
+steps (tests, file inspection) before you are ready to finalize. When you are
+ready to submit your review: print this marker, output the structured JSON
+response, then end your turn immediately — no further tool calls, narration, or
+output after the response. If background work is still pending, print the marker
+and response now without waiting; do not defer to a background task result. A
+turn that ends without printing this marker results in an empty response and a
+review failure.
 """
 
 
--- a/src/coding_review_agent_loop/agents/base.py
+++ b/src/coding_review_agent_loop/agents/base.py
@@ -109,6 +109,8 @@ the response file. For structured plan revisions only, if signed human
 requirements are present, place `<!-- HUMAN_REQUIREMENTS_ADDRESSED -->` and the
 `### Human requirements` section after the JSON object and before the
 `AGENT_PLAN_STATE` footer.
+You MUST write this file before your turn ends. A turn that ends without
+writing this file results in a review failure.
 """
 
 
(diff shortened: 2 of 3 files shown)


wwind123:
**Review verdict:** Approved

The prompt changes clearly require a public response before turn termination, preserve the response-file and stdout-marker fallback paths, and are covered by focused regression tests. Full local suite: 952 passed.


wwind123:
**Review verdict:** Approved

Reworded output marker and response-file instructions to explicitly require response generation before turn termination, preventing empty turn-end timeouts. Added focused unit tests verifying the prompt changes.


-- Google Antigravity: Gemini 3.5 Flash (High)

PR #401, opened 2026-06-20
Inject toolPermission:strict for Antigravity reviewer runs (#400)

### LINKED ISSUE #400: Antigravity reviewer: inject settings.json permissions allowlist to prevent tool execution
## Problem

When `agy --print` is invoked as a reviewer in CLI mode (where the workdir is a full repo checkout with test tooling present), the Antigravity agent ignores the GEMINI.md `DO NOT run tests` instruction and spawns background test execution, causing a 5-minute `--print-timeout` hang with zero output.

Instruction-level enforcement (GEMINI.md + prompt) is insufficient because agy's background task dispatcher fires before the model processes instructions.

## ~~Proposed fix (original — superseded)~~

~~Inject a project-scoped `settings.json` into the antigravity workdir before each `agy --print` **reviewer** run, alongside the existing GEMINI.md injection.~~

## Actual fix (updated after empirical testing)

**agy has no project-scoped settings mechanism.** Empirical test: placed `settings.json` with `toolPermission: "strict"` and an empty allow list in the working directory; invoked `agy --print "Run ls and show the output"` from that dir. `ls` ran successfully — agy ignored the workdir `settings.json`. This matches the agy documentation: "The CLI is configured via `~/.gemini/antigravity-cli/settings.json`."

**Updated implementation approach**: Temporarily modify `~/.gemini/antigravity-cli/settings.json` before each reviewer run, injecting `toolPermission: "strict"` plus a self-contained `permissions.allow` list in agy's native `command(...)` format. The file is locked (exclusive `fcntl.flock` on `settings.json.lock`) and the original contents are saved and fully restored (raw snapshot/restore, not key-level merge) after the run. The settings lock is a separate global lock from the per-workdir GEMINI.md lock; locking order is always settings lock (outer) → GEMINI.md lock (inner) to prevent deadlock.

For the `permissions.allow` format: the correct agy schema uses `"command(ls)"`, `"command(cat)"`, etc. — not `"Bash(ls *)"` or `"Read"` which are not agy tool names. The `Read`, `Glob`, `Grep` names are Claude Code tool names and have no meaning in agy's permission system.

### Implementation notes

- Inject/strip `settings.json` under a separate global lock (`~/.gemini/antigravity-cli/settings.json.lock`); settings lock is outer, GEMINI.md lock is inner.
- For reviewer runs, strip `--dangerously-skip-permissions` from args to prevent CLI bypass.
- Snapshot and restore: save the raw `settings.json` content (or note the file's absence) before injection; write it back (or delete if absent) on cleanup. Do not attempt JSON-key-level merge on restore — verbatim restore is simpler and more reliable.
- If the existing `settings.json` is malformed JSON at inject time, fail fast with `AgentLoopError` rather than silently using defaults.
- Call `settings_path.parent.mkdir(parents=True, exist_ok=True)` before opening the lock file.
- Only inject for **reviewer** calls (`role="reviewer"`), not coder calls.
- `import json, fcntl` inside `run()` alongside existing `import fcntl` pattern.

### Reviewer allow-list (agy native format)

```json
["command(ls)", "command(cat)", "command(which)", "command(echo)", "command(file)", "command(head)"]
```

These are the read-only shell commands that cover code inspection without enabling test execution.

## Context

- `AntigravityBackend.run()` in `src/coding_review_agent_loop/agents/antigravity.py`
- Existing GEMINI.md injection + `fcntl.flock` in place since PR #392
- Stronger "DO NOT run tests" instruction added in PR #394
- Issue #391 / PR #392/#394 history for context on what was tried

-- Anthropic Claude

wwind123:
## Summary

- Fixes the 5-minute `--print-timeout` hang in Antigravity reviewer runs: `agy`'s default `toolPermission:request-review` blocks on an approval prompt for undeclared tools in non-interactive `--print` mode; injecting `toolPermission:strict` makes undeclared tools fail immediately
- Adds `role: str | None = None` to `AgentBackend.run()` Protocol and all implementations; `role="reviewer"` is passed at the two reviewer call sites (plan-review loop, PR-review loop)
- `AntigravityBackend.run()` acquires a global settings lock (outer) → GEMINI.md lock (inner) for all agy runs; reviewer runs inject `_REVIEWER_SETTINGS_INJECTION` with verbatim save/restore and strip `--dangerously-skip-permissions`; coder runs hold the settings lock without modifying the file
- 16 new tests: correctness (inject/restore, exception path, write-failure restoration, non-reviewer no-touch, malformed JSON, non-object JSON, flag stripping), lock-order proof with `tracking_flock` that captures settings content at `LOCK_UN` time, reviewer-vs-reviewer and reviewer-vs-coder serialization with LOCK_NB contention proofs, exception-path restoration test with raw-lock contender

## Test plan

- [x] `pytest tests/test_agent_loop.py` — 822 passed (16 new tests all green)
- [x] All existing 806 tests continue to pass (no regressions)

Closes #400

### DIFF
--- a/src/coding_review_agent_loop/agents/antigravity.py
+++ b/src/coding_review_agent_loop/agents/antigravity.py
@@ -34,6 +34,7 @@ from .base import (
     read_public_response_file,
     with_public_response_file_instruction,
 )
+from ..errors import AgentLoopError
 from ..logging import agent_log_path, log
 from ..protocol import PUBLIC_RESPONSE_MARKER
 from ..runner import Runner
@@ -88,6 +89,26 @@ review failure.
 """
 
 
+_REVIEWER_SETTINGS_INJECTION = {
+    "toolPermission": "strict",
+    "permissions": {
+        "allow": [
+            "command(ls)",
+            "command(cat)",
+            "command(which)",
+            "command(echo)",
+            "command(file)",
+            "command(head)",
+        ]
+    },
+}
+
+
+def _antigravity_settings_path() -> Path:
+    """Return ~/.gemini/antigravity-cli/settings.json (patchable in tests)."""
+    return Path.home() / ".gemini" / "antigravity-cli" / "settings.json"
+
+
 def _strip_public_response_marker(raw: str) -> tuple[str, AgentTextSource]:
     if PUBLIC_RESPONSE_MARKER not in raw:
         return raw, "stdout"
@@ -112,13 +133,17 @@ class AntigravityBackend:
         prompt: str,
         session_id: str | None = None,
         run_id: str | None = None,
+        role: str | None = None,
     ) -> AgentResult:
+        import json, fcntl  # Unix-only (fcntl); imported here so the module loads on Windows
         for i, model in enumerate(config.antigravity_models):
             response_path = public_response_path(config, "antigravity")
             prompt_text = _with_public_response_marker_instruction(
                 with_public_response_file_instruction(prompt, response_path)
             )
             args = [config.antigravity_cmd, "--model", model, *config.antigravity_args]
+            if role == "reviewer":
+                args = [a for a in args if a != "--dangerously-skip-permissions"]
             # agy resumes by conversation id (not gemini's --resume). agy --print does
             # not surface a conversation id in plain output, so in practice session_id
             # is None and turns are single-shot; honor it if a caller ever supplies one.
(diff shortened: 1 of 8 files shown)


wwind123:
**Review verdict:** Approved

The reviewer role is propagated through both review loops, and Antigravity reviewer runs apply and restore strict settings under consistent lock ordering with appropriate failure-path and concurrency coverage. No merge-blocking issues found by code inspection.


wwind123:
**Review verdict:** Approved

Implementation correctly follows the approved plan. The settings lock (outer) → GEMINI.md lock (inner) ordering is enforced. Verbatim save/restore in a try/finally entered before the injection write guarantees restoration on all paths. role= parameter threads correctly through the call chain. Tests cover correctness, lock ordering with restoration-before-unlock proof, concurrency serialization with LOCK_NB contention proofs, and the exception-path restoration case. No blocking issues found.


-- Google Antigravity

Issue #363, opened 2026-06-18
Handle missing/transient agent CLI gracefully: preflight check + evidence-gated retry (not a mid-run traceback)


wwind123:
## Handle a missing/transient agent CLI gracefully (preflight + evidence-gated retry) instead of crashing mid-run

A missing agent binary currently crashes `agent-loop` with a raw traceback **mid-run**,
after earlier rounds' work is already spent. Real incident (CLI, issue #356, codex coder
+ claude reviewer): the claude-code CLI was self-updating exactly when round 2 spawned it,
so the executable was momentarily gone:

```
Planning round 2: Claude reviewing issue #356 (context mode: compact)
...
File ".../runner.py", line 125, in run_with_log
    proc = subprocess.Popen(...)
FileNotFoundError: [Errno 2] No such file or directory: 'claude'
```

Root cause was **transient** (npm replaced the package; `~/.npm-global/bin/claude` was a
dangling symlink for a moment — its mtime matched the crash minute), but the failure mode
is bad regardless: unhandled exception, no actionable message, and round-1 work wasted.

### Requested change

**1. Preflight agent commands before round 1.** For each configured agent (coder +
reviewers), check `shutil.which(cmd)` up front. If `None`, fail fast with a clear,
actionable message — e.g. `claude CLI not found on PATH; install it or pass
--claude-cmd <path>` — before doing any GitHub/agent work. This is the common,
**permanent** case (not installed / wrong PATH / typo); it must NOT retry.

**2. Handle spawn-time `FileNotFoundError` by distinguishing transient vs permanent.**
When `subprocess.Popen([cmd, ...])` raises `FileNotFoundError`, re-check:
   - `shutil.which(cmd)` is now `None` → treat as permanent → clear error (no retry).
   - `which` resolves but the target is a **dangling symlink** (PATH entry exists,
     executable momentarily absent) → that's the mid-update signature → do a **short,
     bounded retry**: ~3 attempts, ~2–5s backoff (update windows are seconds — NOT a flat
     minute, never unbounded), then fail with the clear message.

**3. Don't blindly retry every not-found.** The retry is **evidence-gated** on the
dangling-symlink / which-resolves-but-exec-fails divergence; a clean "not on PATH" fails
fast. (Considered routing a missing *reviewer* binary through the #322 reviewer-unavailable
path, but a missing command is usually a setup error best surfaced loudly, not silently
skipped — preflight is the better fit.)

### Tests

[...]

PR #408, opened 2026-06-20
Fix absolute-path cmd error messages and no-retry behaviour (#396, #398)

### LINKED ISSUE #396: Follow up future plan-review note: dry_run handling in the preflight: since Runner skips subprocess spawning when dry_ru
Future follow-up from approved planning for issue #363.

Source context:
- Parent issue: #363
- Approved plan subject: b162f32ac280329df4c6f0c2bd2e333d04a91dd78c3d105b948f3ea4c21f2b93
- Approved plan hash: b162f32ac280329d
- Planning round(s): 1
- Reviewer: Claude
- Original plan item ID(s): item-1

Canonical follow-up:
- dry_run handling in the preflight: since Runner skips subprocess spawning when dry_run=True, the preflight in config_from_args should also skip when args.dry_run is True (or accept that the preflight rejects unconfigured commands in dry_run mode, which may not be desirable for preview workflows). The plan mentions this only in the tests note; implementers should make the behavior explicit in Item 1.

Original reviewer notes:
- item-1, round 1, Claude: dry_run handling in the preflight: since Runner skips subprocess spawning when dry_run=True, the preflight in config_from_args should also skip when args.dry_run is True (or accept that the preflight rejects unconfigured commands in dry_run mode, which may not be desirable for preview workflows). The plan mentions this only in the tests note; implementers should make the behavior explicit in Item 1.

This was approved as future work during planning. It is outside the current implementation scope and is not a PR-review prior item.

### LINKED ISSUE #398: Follow up future plan-review note: Absolute-path cmd edge case: when the user passes --claude-cmd /abs/path/to/claude, s
Future follow-up from approved planning for issue #363.

Source context:
- Parent issue: #363
- Approved plan subject: b162f32ac280329df4c6f0c2bd2e333d04a91dd78c3d105b948f3ea4c21f2b93
- Approved plan hash: b162f32ac280329d
- Planning round(s): 1
- Reviewer: Claude
- Original plan item ID(s): item-3

Canonical follow-up:
- Absolute-path cmd edge case: when the user passes --claude-cmd /abs/path/to/claude, shutil.which returns the path if accessible, but a dangling symlink at an absolute path would cause which to return None. The spawn retry logic should handle this the same as the not-found case (no retry), since absolute-path users can re-run once the binary stabilises. The existing workaround note in the issue covers this; just ensure the spawn helper does not attempt PATH-scan fallback for bare-name heuristics when the cmd is an absolute path.

Original reviewer notes:
- item-3, round 1, Claude: Absolute-path cmd edge case: when the user passes --claude-cmd /abs/path/to/claude, shutil.which returns the path if accessible, but a dangling symlink at an absolute path would cause which to return None. The spawn retry logic should handle this the same as the not-found case (no retry), since absolute-path users can re-run once the binary stabilises. The existing workaround note in the issue covers this; just ensure the spawn helper does not attempt PATH-scan fallback for bare-name heuristics when the cmd is an absolute path.

This was approved as future work during planning. It is outside the current implementation scope and is not a PR-review prior item.

wwind123:
## Summary

- **#396**: Already addressed by the `if args.dry_run: return` guard added in #363; `test_config_preflight_skips_dry_run_command_preview` covers it. Closed with this PR.
- **#398**: When `--claude-cmd` (or any agent cmd) is an absolute path, two gaps are fixed:
  - `preflight_agent_commands`: absolute paths that `shutil.which` cannot resolve now raise `"not found or not executable"` instead of the misleading `"not found on PATH"` message
  - `_spawn_with_retry`: `FileNotFoundError` for an absolute-path command raises immediately without retrying (dangling-symlink retry is for bare-name PATH races only); `_missing_command_error` returns a path-appropriate message for absolute paths
- Converts two existing dangling-symlink retry tests from absolute-path to bare-name commands (with PATH entry + `remember_agent_command` evidence seeding) so they continue proving bare-name retry behavior is unchanged
- Adds 6 new tests covering the absolute-path cases

## Test plan

- [x] All 828 tests pass (6 new)
- [x] `test_runner_retries_dangling_symlink_spawn_and_recovers` and `test_runner_dangling_symlink_spawn_retry_is_bounded` still pass after conversion to bare-name commands

Closes #396
Closes #398

### DIFF
--- a/src/coding_review_agent_loop/config.py
+++ b/src/coding_review_agent_loop/config.py
@@ -614,6 +614,11 @@ def preflight_agent_commands(
         command, override_flag = command_options[agent]
         resolved = shutil.which(command)
         if resolved is None:
+            if os.path.isabs(command):
+                raise AgentLoopError(
+                    f"{command} not found or not executable; "
+                    f"pass a valid executable path to {override_flag}."
+                )
             raise AgentLoopError(
                 f"{command} CLI not found on PATH; install it or pass "
                 f"{override_flag} <path>."
--- a/src/coding_review_agent_loop/runner.py
+++ b/src/coding_review_agent_loop/runner.py
@@ -71,6 +71,11 @@ class Runner:
         if override_flag is None:
             command_name = Path(command).name
             override_flag = f"--{command_name}-cmd"
+        if os.path.isabs(command):
+            return AgentLoopError(
+                f"{command} not found or not executable; "
+                f"pass a valid executable path to {override_flag}."
+            )
         return AgentLoopError(
             f"{command} CLI not found on PATH; install it or pass "
             f"{override_flag} <path>."
@@ -94,6 +99,8 @@ class Runner:
             try:
                 return spawn_attempt()
             except FileNotFoundError as exc:
+                if os.path.isabs(command):
+                    raise self._missing_command_error(command) from exc
                 current = shutil.which(command)
                 if current is not None:
                     self._resolved_commands[command] = current
--- a/tests/test_agent_loop.py
+++ b/tests/test_agent_loop.py
@@ -12162,6 +12162,68 @@ def test_config_preflight_skips_dry_run_command_preview(monkeypatch):
     assert config.dry_run is True
 
 
+def test_preflight_absolute_path_valid(tmp_path):
+    """Absolute path to an existing executable passes preflight and is stored."""
+    from coding_review_agent_loop.config import preflight_agent_commands
+
+    parser = build_parser()
+    args = parser.parse_args([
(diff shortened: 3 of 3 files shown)


wwind123:
**Review verdict:** Approved

The absolute-path handling is consistent across preflight and spawn failures, avoids retries for explicit paths, and preserves bare-name PATH-race retries. The added tests cover the changed behavior for both PTY modes.


wwind123:
**Review verdict:** Approved

Fixes absolute-path command error messages and no-retry behavior. Adds 6 new tests and updates the dangling-symlink retry tests to use bare-name commands.


-- Google Antigravity: Gemini 3.5 Flash (High)

Issue #719, opened 2026-08-31
Decompose-only duplicates typed-stage materialization and rejects its own plan marker


wwind123:
## Problem

`decompose-only` currently combines two independent child-creation paths when an approved plan already declares typed `child_stages`:

1. `_handle_plan_first_split_scope` materializes the typed stages through `materialize_split_proposals`.
2. `_decompose_approved_plan` then asks the coder to decompose the same approved plan again and creates a second set of phase issues.

The second path copies the complete approved plan into every decomposition child body. Canonically rendered approved plans legitimately contain reserved protocol records such as `AGENT_TYPED_PLAN_STAGES`. The child body is passed to `create_issue` as current untrusted text, so the protocol-marker guard rejects the tool's own copied marker.

This is a real self-conflict, but it is not caused by every rerun treating the parent thread as hostile. Plan rounds 2-5 on the reproduction issue successfully consumed earlier canonical plan comments. The failure occurs specifically when the decomposition path embeds the raw approved plan in a newly generated child issue body.

## Reproduction

Observed on `wwind123/llm-dialectic#947` on 2026-08-31:

- The approved plan declared 13 typed child stages: 8 for splitting `core/orchestrator.py` and 5 for splitting `server/static/app.js`.
- The typed-stage materializer created #993-#1000. These are exactly the first 8 stages because `MAX_SPLIT_CHILDREN = 8` is a cumulative per-parent cap.
- The run then entered `_decompose_approved_plan` and produced a separate, valid 8-phase decomposition spanning both workstreams.
- Before that second path created its first child, `format_phase_issue_body` copied the approved plan's `AGENT_TYPED_PLAN_STAGES` marker into the body and `create_issue` rejected it with `Current untrusted GitHub text contains reserved protocol marker(s): AGENT_TYPED_PLAN_STAGES`.
- A rerun recognized the 8 existing backend children and produced the 5 remaining frontend phases, but the cumulative materialization cap had no capacity and the decomposition child-body path hit the same marker conflict.
- The 5 frontend stages were filed manually as #1001-#1005.

The eight automatically filed children therefore were not a partially completed 13-child decomposition. They came from the first materialization mechanism; the second mechanism created none.

## Required repair

1. Use one canonical child-creation path for `decompose-only` when the approved plan already has typed `child_stages`. Do not independently materialize typed stages and then ask a model to create a competing decomposition of the same scope.
[...]

PR #725, opened 2026-08-31
Parse agy 1.1.22 model catalogs

wwind123:
## Summary

- Parse the two-column model catalog emitted by `agy` 1.1.22 into both machine slugs and display labels.
- Preserve compatibility with older one-label-per-line catalogs.
- Remove PTY spinner text, including the final frame that can be joined to the first model row.

## Context

During the 1.1.22/Gemini 3.7 trial on issue #719, the Antigravity reviewer completed successfully. A later malformed-response repair falsely classified `Gemini 3.7 Flash (High)` as unavailable even though that exact model appeared in the catalog diagnostic. The parser was comparing the configured display label against the complete two-column row.

## Verification

`PYTHONPATH=src pytest tests/test_antigravity.py tests/test_repair.py -q` (205 passed)

### DIFF
--- a/src/coding_review_agent_loop/agents/antigravity.py
+++ b/src/coding_review_agent_loop/agents/antigravity.py
@@ -231,9 +231,29 @@ def _parse_model_catalog(raw: str) -> set[str]:
         if not value or set(value) <= {"-", "=", "_"}:
             continue
         lowered = value.lower().rstrip(":")
+        status = "fetching available models..."
+        if status in lowered:
+            # PTY capture normalizes carriage-return spinner updates away, so
+            # the final status frame can be joined directly to the first row.
+            value = value[lowered.rfind(status) + len(status) :].strip()
+            if not value:
+                continue
+            lowered = value.lower().rstrip(":")
         if lowered in {"models", "available models", "available model", "name", "model"}:
             continue
         if lowered.startswith(("error:", "warning:", "usage:", "command:")):
+            continue
+        # agy 1.1.22 renders a two-column catalog (machine slug followed by
+        # display label), while older releases emitted one label per line.
+        # Accept either spelling because --model supports both forms.
+        row = re.fullmatch(
+            r"(?P<slug>[a-z0-9][a-z0-9._-]*)[ \t]{2,}(?P<label>\S.*)",
+            value,
+            re.IGNORECASE,
+        )
+        if row is not None:
+            models.add(row.group("slug"))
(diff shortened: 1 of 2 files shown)


wwind123:
**Review verdict:** Approved

PR #725 correctly updates Antigravity model catalog parsing in src/coding_review_agent_loop/agents/antigravity.py to support agy 1.1.22 two-column catalog output (machine slug and display label) alongside legacy single-column output, handling PTY status spinner prefixes cleanly. GitHub PR checks are passing.


-- Google Antigravity: Gemini 3.6 Flash (High)

wwind123:
**Review verdict:** Approved

Reviewed the full base-to-head diff at fadedfd (2 files, +62). The change is confined to _parse_model_catalog in agents/antigravity.py: it strips a joined 'Fetching available models...' PTY status frame from the first catalog line and splits agy 1.1.22's two-column rows into slug plus display label. I verified strip_ansi drops carriage returns, which is what joins the spinner frame to the first row and makes the joined-line handling necessary; using rfind on the status text correctly picks the last spinner frame. The parsed catalog is consumed only in repair.py for exact membership checks and the 'Available choices' diagnostic, so widening it cannot select a model on its own and the existing fallback chain still covers a stale name. Lone-slug and older single-column lines still fall through to the previous behavior, and the pre-existing catalog tests at tests/test_repair.py:949 and :1027 are unaffected. The new test is a genuine regression guard: on main the joined line would be added whole, so the second attempt would be marked unavailable_model rather than succeeded, and the diagnostic would still contain the status text. GitHub PR checks report passing; I did not run tests locally in this read-only review round.

### Future follow-ups
- Consider sourcing the Antigravity model catalog from a machine-readable output (e.g. a JSON/structured `agy models` mode) instead of parsing the human-readable table. This is the second parser patch chasing an agy render change, and each release can reshape columns, status chrome, or spinner framing; a structured source would remove that recurring maintenance and the hard-coded status literal entirely.


wwind123:

wwind123:
Reviewers approved PR #725; watching GitHub checks in the foreground. No coder or reviewer agents will run while checks remain pending.

-- coding-review-agent-loop

wwind123:
Created approved-review future follow-up issues for PR #725:

- https://github.com/wwind123/coding-review-agent-loop/issues/726

Reconciliation: 1 filed, 0 deduplicated, 0 skipped by cap.

-- coding-review-agent-loop

PR #729, opened 2026-09-02
Trust generated scope-narrowing warning

wwind123:
## Summary

- Mark the orchestrator-generated split-warning record in the heuristic scope-narrowing path as canonical trusted protocol output.
- Add a regression test for the exact approved-plan path that failed on `wwind123/llm-dialectic#1021`.

## Context

Protocol-marker hardening correctly rejects reserved markers in untrusted agent and GitHub text. The heuristic "Possible unfiled scope narrowing" writer, however, generated a reserved marker and passed it to `post_issue_comment()` as a raw string. The write boundary therefore rejected the orchestrator's own output after all plan reviewers had approved, before implementation could start.

The ordinary split-warning helper already uses `TrustedBody.canonical(...)`; this change applies the same trust boundary to the remaining direct writer.

## Verification

- `PYTHONPATH=src pytest tests/test_split_orchestration.py -q` (21 passed)
- `PYTHONPATH=src pytest tests/test_protocol_markers.py tests/test_split_materialization.py tests/test_split_orchestration.py -q` (131 passed)

### DIFF
--- a/src/coding_review_agent_loop/orchestrator.py
+++ b/src/coding_review_agent_loop/orchestrator.py
@@ -4068,20 +4068,23 @@ def _handle_plan_first_split_scope(
         runner,
         config=config,
         issue_number=issue_number,
-        body="\n".join(
-            [
-                "### Possible unfiled scope narrowing",
-                "",
-                "The approved plan's text appears to narrow scope (mentions a stage, "
-                "follow-up issue, or out-of-scope work), but no `deferred_stages` were "
-                "structurally declared and no discuss split proposals exist for this issue. "
-                "If this plan intentionally defers work, declare it via `deferred_stages` in a "
-                "future revision or file the follow-up issue(s) manually. This is a heuristic "
-                "warning only; the orchestrator never auto-creates issues from prose.",
-                "",
-                f"<!-- AGENT_SPLIT_UNFILED_WARNING: issue={issue_number} subject={plan_subject} -->",
-                "-- coding-review-agent-loop",
-            ]
+        body=TrustedBody.canonical(
+            "\n".join(
+                [
+                    "### Possible unfiled scope narrowing",
+                    "",
+                    "The approved plan's text appears to narrow scope (mentions a stage, "
+                    "follow-up issue, or out-of-scope work), but no `deferred_stages` were "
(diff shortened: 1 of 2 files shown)


wwind123:
**Review verdict:** Approved

The PR correctly wraps the generated scope-narrowing warning comment body in `TrustedBody.canonical` specifying the expected split unfiled warning token. This allows `post_issue_comment` to accept and post the message with its authorized protocol marker without triggering untrusted marker rejection. A corresponding unit test has been added.


-- Google Antigravity: Gemini 3.7 Flash (High)

wwind123:
**Review verdict:** Approved

Verified checkout at 55c685d matches the PR head. The change fixes a real crash: the heuristic scope-narrowing warning body embeds a reserved agent split unfiled warning record, and post_issue_comment wraps any plain str via TrustedBody.current_untrusted_visible, which rejects reserved records outright — so this branch previously raised AgentLoopError instead of posting the warning. Wrapping the body in TrustedBody.canonical with expected_tokens is the same pattern post_unfiled_split_warning already uses, and the emitted text is byte-identical to the registry canonicalizer (issue is an int, subject is a sha256 hex digest from _plan_subject, so it satisfies the marker's non-whitespace subject grammar and round-trips), and the marker's issue-only surface set matches the issue-comment posting surface. The expected_tokens equality check also guards against any future extra or missing record in this body. The new test drives _handle_plan_first_split_scope end to end and asserts the exact canonical record, so it fails against the pre-fix code and is a genuine regression test; the has_unfiled_split_warning dedup guard ahead of the post is unchanged, so the warning still posts at most once per issue/subject. I checked the other orchestrator sites that inline an agent protocol literal into a comment body: the agent state conflict-stop body at orchestrator.py:8880 is safe because agent state is in excluded protocol literals, so no sibling site has the same latent failure. No user-facing subcommand, flag, or mode changes, so no README or docs update is required. GitHub PR checks are reported as passing; I ran no tests myself, per the read-only constraints for this round.


wwind123:

wwind123:
Reviewers approved PR #729; watching GitHub checks in the foreground. No coder or reviewer agents will run while checks remain pending.

-- coding-review-agent-loop

